PR #33, opened 2022-04-18
Doc updates

lazovicff:


lazovicff:
Docs:
https://docs.rs/eigen-trust/0.1.0/eigen_trust/

Crates:
https://crates.io/crates/eigen-trust/0.1.0

PR #52, opened 2022-05-20
Readme update

lazovicff:


lazovicff:
https://crates.io/crates/eigen-trust

PR #81, opened 2022-09-05
And circuit tests and inline documentation

4rmanc:


circuit/src/gadgets/and.rs
```diff
@@ -190,4 +203,43 @@ mod test {
...
+	#[test]
+	fn test_and_x1_y0() {
+		// Testing x = 1 and y = 0
+		let test_chip = TestCircuit::new(Fr::from(1), Fr::from(0));
+
+		let k = 4;
+		let rng = &mut rand::thread_rng();
+		let params = generate_params(k);
+		let res =
+			prove_and_verify::<Bn256, _, _>(params, test_chip, &[&[Fr::from(0)]], rng).unwrap();
```
lazovicff, line 216: We should use `MockProver::run` for every test.

circuit/src/gadgets/and.rs
```diff
@@ -21,6 +21,17 @@ pub struct AndChip<F: FieldExt> {
...
+	/// # Examples
+	///
+	/// ```
+	/// let x = 1;
+	/// let y = 0;
+	/// // (x * y) - z == 0
+	/// //z is the next rotation cell for the x value.
+	/// let z = (x * y);
+	/// z;
+	/// ```
```
lazovicff, line 34: We can move this example to the [line 56](https://github.com/lazovicff/eigen-trust/pull/81/files#diff-44a851e33abd6b4304994985eb6bc8d63d0154e82f1c43b436b1be6720198e39R56), and replace x and y with the actual values.

lazovicff:
It would be useful to note here that values `x` and `y` have to be boolean values: https://github.com/lazovicff/eigen-trust/pull/81/files#diff-44a851e33abd6b4304994985eb6bc8d63d0154e82f1c43b436b1be6720198e39L66-L68

circuit/src/gadgets/and.rs
lazovicff: Yes

lazovicff:
Nice!
There are few more things cargo doc will require from us:
1. Comments above [AndConfig](https://github.com/4rmanc/eigen-trust/blob/and-circuit-tests/circuit/src/gadgets/and.rs#L10-L15) and all of its properties.
2. Comments above [AndChip](https://github.com/4rmanc/eigen-trust/blob/and-circuit-tests/circuit/src/gadgets/and.rs#L18-L21) and all of its properties.
3. Comments above [`new`](https://github.com/4rmanc/eigen-trust/blob/and-circuit-tests/circuit/src/gadgets/and.rs#L24) method.

Use this as a guide: https://doc.rust-lang.org/rust-by-example/meta/doc.html#doc-comments

4rmanc:
Should I comment as you did for configure and synthesize? One sentence, short and fundamental.

lazovicff (approved):

PR #82, opened 2022-09-06
Sum circuit tests and documentation

4rmanc:
For the test case large arrays with random values, I entered 20 same but random elements as input. I thought looping for each input and entering different random numbers but that would be inefficient for the test case. Also, to enter more than 3 inputs I changed the configuration struct as a const S so it would take as many as inputs in an array I sent there. If you did 3 elements on purpose I can change it back.


circuit/src/gadgets/sum.rs
```diff
@@ -0,0 +1,226 @@
...
+		assert_eq!(prover.verify(), Ok(()));
+	}
+
+	#[test]
+	fn test_sum_zero_arr() {
+		// Testing zero array. Returns InvalidInstance error.
+		let test_chip: TestCircuit<Fr, 0> = TestCircuit::new([]);
+
+		let k = 4;
+		let prover = MockProver::run(k, &test_chip, vec![]);
```
lazovicff, line 210: `InvalidInstance` error is given because the instance cell is not populated.
What we want is to test if empty array yields to sum of zero.
So perhaps pass `Fr::zero()` as a public input.

circuit/src/gadgets/sum.rs
```diff
@@ -0,0 +1,226 @@
...
+		let pub_ins = vec![Fr::from(6)];
+		let prover = MockProver::run(k, &test_chip, vec![pub_ins]).unwrap();
+		assert_eq!(prover.verify(), Ok(()));
+	}
+
+	#[test]
+	fn test_sum_random_element() {
+		// Testing big array with a random element from the field.
+		let rng = &mut rand::thread_rng();
+		let array = [<Fr as Field>::random(rng); 20];
```
lazovicff, line 195: This will create 20 of the same field values. We want every value to be different.
And let's try bigger arrays, so perhaps try:
```rust
let array = [(); 512].map(|_| <Fr as Field>::random(rng));
```

4rmanc: But in that case how I can calculate the public input? Should I use different calculators to get the public input and test the circuit?

lazovicff: Yes, loop through the array and calculate the sum, and the result as a public input

circuit/src/gadgets/sum.rs
4rmanc: I see that an empty array doesn't yield to sum of zero. It gives a different error which is because the chip is not satisfied. Are we aiming to get a result from the circuit when we give an empty array?

lazovicff: Yes, there is probably no use-case for it, but we should cover all the edge-cases.

HINT: The selector is enabled at index 0, before we start the loop: https://github.com/lazovicff/eigen-trust/blob/sum-circuit-test-doc/circuit/src/gadgets/sum.rs#L75, maybe that's unnecessary

4rmanc: Oh, okay. If selector never enables it will provide zero because of the initial sum will provide zero as a constant at the beginning. So we won't get an error about chip satisfaction.

lazovicff:
We should remove `accumulate.rs` before merging.

lazovicff (approved):

PR #86, opened 2022-09-08
test and documentation

4rmanc:


circuit/src/gadgets/is_zero.rs
```diff
@@ -41,8 +49,18 @@ impl<F: FieldExt> IsZeroChip<F> {
...
+				// that one of the variables is 0.
 				sel_exp.clone() * (x_exp.clone() * b_exp.clone()),
 				// x * x_inv + b - 1 == 0
+				// Example 1:
+				// If x = 1 => x_inv = 1;
+				// (1 * 1 + b - 1) == 0
+				// In this case our b value must be 0.
+				// Example 2:
+				// If b = 1
+				// (x * x_inv + 0 - 1) == 0 => (x * x_inv) must be equal to 0.
```
lazovicff, line 62: In example 2:
It's stated that b is 1, but in the formula, b is replaced by 0

circuit/src/gadgets/is_zero.rs
```diff
@@ -41,8 +49,18 @@ impl<F: FieldExt> IsZeroChip<F> {
 
 			vec![
 				// x * b == 0
+				// Here we check this constraint because we want to be sure
+				// that one of the variables is 0.
```
lazovicff, line 53: We should state that b is the boolean (x == 0) we want to output

lazovicff (approved):

PR #88, opened 2022-09-09
select test and documentation

4rmanc:
I deleted your [question](https://github.com/lazovicff/eigen-trust/blob/master/circuit/src/gadgets/select.rs#L34) comment. I can change it again at the second milestone phase if it seems more correct.

### DIFF
--- a/circuit/src/gadgets/select.rs
+++ b/circuit/src/gadgets/select.rs
@@ -73,7 +73,7 @@ impl<F: FieldExt> SelectChip<F> {
 				// Example 2:
 				// bit = 0
 				// z is the next rotation cell for the y value.
-				// Next rotation cell for y will carry the same value when bit == 0. (x == z)
+				// Next rotation cell for y will carry the same value when bit == 0. (y == z)
 				// x = 5
 				// y = 3
 				// z = 3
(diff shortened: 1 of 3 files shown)


circuit/src/gadgets/select.rs
```diff
@@ -53,11 +61,23 @@ impl<F: FieldExt> SelectChip<F> {
...
+				// x = 5
+				// y = 3
+				// z = 5
+				// 1 * (x - y) - (z - y) = 1 * (5 - 3) - (5 - 3) = 0
+				// Example 2:
+				// bit = 0
+				// x = 5
+				// y = 3
+				// z = 5
+				// 0 * (x - y) - (z - y) = 0 * (5 - 3) - (5 - 3) = 2
```
lazovicff, line 78: This does not evaluate to 0 because of the assumptions made above

circuit/src/gadgets/select.rs
```diff
@@ -86,11 +106,14 @@ impl<F: FieldExt> SelectChip<F> {
 			|mut region: Region<'_, F>| {
 				config.selector.enable(&mut region, 0)?;
 
+				// Here we check bit is boolean or not.
```
lazovicff, line 109: This is not the place where we do this check. We are checking it here: https://github.com/lazovicff/eigen-trust/blob/select-circuit-test-doc/circuit/src/gadgets/select.rs#L101-L102

circuit/src/gadgets/select.rs
4rmanc: I think you meant it does evaluate to 0. I see now that our z value is also the selected result value here. It makes constraint == 0 and returns the selected variable.

lazovicff (approved):

PR #90, opened 2022-09-12
tests and docs

4rmanc:
At this [line](https://github.com/lazovicff/eigen-trust/blob/lt-eq-test-doc/circuit/src/gadgets/lt_eq.rs#L98) I tried to emphasize that we use bits2num and checking the result. It can be a little bit low-level if someone just reads that I think. What is your opinion should I change that and explain it in more detail?


circuit/src/gadgets/lt_eq.rs
```diff
@@ -73,7 +84,20 @@ impl<F: FieldExt> LessEqualChip<F> {
 
 			let x_next_exp = v_cells.query_advice(x, Rotation::next());
 
-			vec![s_exp * ((x_exp + n_shifted_exp - y_exp) - x_next_exp)]
+			vec![
+				// (x + n_shifted - y) - z == 0
+				// n_shifted value is equal to smallest 253 bit number.
+				// Because of that calculations will be done in between 252 to 254 bit range.
+				// That range can hold 252 bit number calculations without overflowing.
```
lazovicff, line 91:
```suggestion
				// Because of that calculations will be done in between the 252 to 254-bit range.
				// That range can hold 252-bit number calculations without overflowing.
```

lazovicff:
I think comments are good enough, they are clear and understandable. Just some fixes I got from Grammarly

4rmanc:
I found some other grammar mistakes from old commitments. Can I push these changes to this branch as well?

lazovicff:
@4rmanc Yes, that would be great

lazovicff (approved):

PR #91, opened 2022-09-16
documentation

4rmanc:
Is it good to refer to a [link](https://github.com/lazovicff/eigen-trust/blob/poseidon-doc/circuit/src/poseidon/mod.rs#L321) like this?


circuit/src/poseidon/mod.rs
```diff
@@ -20,6 +21,7 @@ pub struct PoseidonConfig<const WIDTH: usize> {
 	partial_round_selector: Selector,
```
lazovicff, line 21: We also have to document these fields.
Same goes for every struct.

circuit/src/poseidon/params/mod.rs
```diff
@@ -4,22 +4,27 @@ pub mod bn254_5x5;
 use halo2wrong::halo2::{arithmetic::FieldExt, plonk::Expression};
 
 pub trait RoundParams<F: FieldExt, const WIDTH: usize>: Sbox {
+	/// Returns integer 8 as usize.
```
lazovicff, line 7: This is just a trait definitions.
For every parameter set there will be a different implementation.
So, let's just say that this function should return a number of full rounds.

circuit/src/poseidon/params/mod.rs
```diff
@@ -4,22 +4,27 @@ pub mod bn254_5x5;
...
 	fn full_rounds() -> usize;
+	/// Returns integer 60 as usize.
```
lazovicff, line 9: Same here, let's say that this returns a number of partial rounds.

circuit/src/poseidon/params/mod.rs
```diff
@@ -4,22 +4,27 @@ pub mod bn254_5x5;
...
 
+	/// Returns total count size.
 	fn round_constants_count() -> usize {
 		let partial_rounds = Self::partial_rounds();
 		let full_rounds = Self::full_rounds();
 		(partial_rounds + full_rounds) * WIDTH
 	}
 
+	/// Returns enough length vector to track counting.
 	fn round_constants() -> Vec<F> {
```
lazovicff, line 20: This comment is not clear. This function returns the round constants array to be used in permutation.

circuit/src/poseidon/params/mod.rs
```diff
@@ -28,20 +33,26 @@ pub trait RoundParams<F: FieldExt, const WIDTH: usize>: Sbox {
 		result
 	}
 
+	/// Returns fixed string array values as field elements.
 	fn mds() -> [[F; WIDTH]; WIDTH] {
```
lazovicff, line 37: This function returns MDS matrix with a size of WIDTH x WIDTH

circuit/src/poseidon/params/mod.rs
```diff
@@ -28,20 +33,26 @@ pub trait RoundParams<F: FieldExt, const WIDTH: usize>: Sbox {
...
 		let mds_raw = Self::mds_raw();
 		mds_raw.map(|row| row.map(|item| hex_to_field(item)))
 	}
 
+	/// Returns fixed string vector.
```
lazovicff, line 42: This function returns rounds constants in its hex string form.

circuit/src/poseidon/params/mod.rs
```diff
@@ -28,20 +33,26 @@ pub trait RoundParams<F: FieldExt, const WIDTH: usize>: Sbox {
...
 	fn round_constants_raw() -> Vec<&'static str>;
+	/// Returns fixed string array.
```
lazovicff, line 44: This functions returns MDS matrix in its hex string form.

circuit/src/poseidon/params/mod.rs
```diff
@@ -28,20 +33,26 @@ pub trait RoundParams<F: FieldExt, const WIDTH: usize>: Sbox {
...
+	/// Returns round constants in its hex string form.
 	fn round_constants_raw() -> Vec<&'static str>;
+	/// Returns MDS martrix in its hex string form.
 	fn mds_raw() -> [[&'static str; WIDTH]; WIDTH];
 }
 
 pub trait Sbox {
+	/// Returns exp^5.
 	fn sbox_expr<F: FieldExt>(exp: Expression<F>) -> Expression<F>;
+	/// Returns f^5.
```
lazovicff, line 51: These should just say that it applies the Sbox exponentiation. The exponentiation is different for each implementation

lazovicff (approved):

PR #100, opened 2022-09-22
bits2num documentation

4rmanc:


circuit/src/gadgets/bits2num.rs
```diff
@@ -158,7 +185,7 @@ mod test {
 				},
 			)?;
 
-			let bits = to_bits::<256>(self.numba.to_bytes()).map(|b| Fr::from(b));
+			let bits = to_bits::<256, N>(self.bytes).map(|b| Fr::from(b));
 			let bits2num = Bits2NumChip::new(numba, bits);
```
lazovicff, line 189: This `bits` array will hold 256 bits of zeros, and we are passing that to the Bits2Num chip. Our goal was to pass the empty array

4rmanc: Carrying if-else control flow to the synthesize worked. Is this the optimal solution?

lazovicff (approved):

PR #104, opened 2022-09-26
Common gadget

4rmanc:
Minimized the rows to a single row for better performance at the prover side (Except select circuit). For the test cases, implemented the [match control flow](https://github.com/eigen-trust/eigen-trust/blob/common-gadget/circuit/src/gadgets/common.rs#L422-L476). Every gadget we put into common gadget have different starting letter, which is used in this flow to differentiate them. Also, normally for the is_bool circuit, we don't need to check the public instance but in this test circuit design, I put dummy inputs to the public instances for the is_bool [test cases](https://github.com/eigen-trust/eigen-trust/blob/common-gadget/circuit/src/gadgets/common.rs#L540-L586) and they don't affect the circuit.

### DIFF
--- a/circuit/src/gadgets/common.rs
+++ b/circuit/src/gadgets/common.rs
@@ -1,0 +1,751 @@
+use halo2wrong::halo2::{
+	arithmetic::FieldExt,
+	circuit::{AssignedCell, Layouter, Region, Value},
+	plonk::{Advice, Column, ConstraintSystem, Error, Expression, Selector},
+	poly::Rotation,
+};
+use std::marker::PhantomData;
+
+#[derive(Copy, Clone, Debug)]
+/// Configuration elements for the circuit are defined here.
+pub struct CommonConfig {
+	/// Configures columns for the advice.
+	advice: [Column<Advice>; 3],
+	/// Configures fixed boolean values for each row of the circuit.
+	selectors: [Selector; 6],
+}
+
+/// Structure for the chip.
+pub struct CommonChip<F: FieldExt> {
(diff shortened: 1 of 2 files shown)


circuit/src/gadgets/common.rs
```diff
@@ -0,0 +1,751 @@
...
+			}
+
+			match self.c {
+				'a' => {
+					let and = CommonChip::and(
+						items[0].clone(),
+						items[1].clone(),
+						config.common,
+						layouter.namespace(|| "and"),
+					)?;
```
lazovicff, line 429: Why did you decide to use a char to match, instead of using an `enum`. e.g.:
```rust
enum Gadgets {
  ...
  And(F, F),
  IsBool(F),
  ...
}
```

4rmanc: In your example, I need to carry [assign_region](https://github.com/eigen-trust/eigen-trust/blob/common-gadget/circuit/src/gadgets/common.rs#L408) inside each match case to access enums data if I am not missing anything at rust knowledge side.
Maybe  `enum Gadgets{
		And,
		IsBool,
		IsEqual,
		IsZero,
		Mul,
		Select
	}` will be more aesthetic.
In this case input will be something like this: `let test_chip = TestCircuit::new([Fr::from(1), Fr::from(1)], Gadgets::And);`
I used char in a similar manner, now I think the enum variant seems more aesthetic.

lazovicff: Ah, right. That solution sounds good

lazovicff (approved):

PR #106, opened 2022-09-30
eddsa-common-optimization

4rmanc:
Carried `eddsa` gadgets' configuration gates to the [common](https://github.com/eigen-trust/eigen-trust/blob/eddsa-common/circuit/src/gadgets/common.rs) and test cases to the [mod](https://github.com/eigen-trust/eigen-trust/blob/eddsa-common/circuit/src/eddsa/mod.rs). I implemented the common circuit to all of the other files so that we could delete files after the last task. Also, [here](https://github.com/eigen-trust/eigen-trust/blob/eddsa-common/circuit/src/gadgets/common.rs#L195) I did hard coding for the bits2num configuration since we always use 256 bits.

### DIFF
--- a/circuit/src/eddsa/eddsa_gadgets.rs
+++ b/circuit/src/eddsa/eddsa_gadgets.rs
@@ -1,0 +1,646 @@
+use super::native::ops::{add_exp, add_value, double_exp, double_value};
+use crate::gadgets::bits2num::{Bits2NumChip, Bits2NumConfig};
+use halo2wrong::{
+	curves::bn256::Fr,
+	halo2::{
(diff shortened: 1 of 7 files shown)


lazovicff:
Did you consider having separate chip for Eddsa gadgets? It would be more contextually correct to put it into `eddsa` folder.

4rmanc:
I did but the task description "**combine with the common gadget**" got me confused a bit. Should I create a new file and a chip in the `eddsa` folder or a chip inside the `common` file?

lazovicff:
@4rmanc I will leave that to you to decide

lazovicff (approved):

PR #111, opened 2022-11-04
Non native arith

lazovicff:


circuit/src/wrong/native.rs
```diff
@@ -0,0 +1,371 @@
...
+		let mut big_answer = BigUint::one();
+		for i in 0..8 {
+			big_answer *= a_big_array[i].clone();
+			let a = Integer::<Fq, Fr, 4, 68, Bn256_4_68>::new(a_big_array[i].clone());
+			acc = acc.result.mul(&a);
+		}
+
+		assert!(Bn256_4_68::constrain_binary_crt(
+			acc.intermediate, acc.result.limbs, acc.residues
+		));
```
lazovicff, line 216: We should check constraints at each iteration.
Apply to all other tests

circuit/src/wrong/native.rs
```diff
@@ -0,0 +1,371 @@
...
+impl<W: FieldExt, N: FieldExt, const NUM_LIMBS: usize, const NUM_BITS: usize, P>
+	Quotient<W, N, NUM_LIMBS, NUM_BITS, P>
+where
+	P: RnsParams<W, N, NUM_LIMBS, NUM_BITS>,
+{
+	/// Returns Quotient type for the addition.
+	pub fn add(self) -> N {
+		match self {
+			Quotient::Add(res) => res,
+			_ => Option::<N>::None.unwrap(),
```
lazovicff, line 27: Why not just return Option and unwrap it when you want to use it

4rmanc: Do you mean returning Some(Quotient::Add(self)).unwrap()? I thought returning a value would not be good for exception handling

lazovicff:
```rust
pub fn add(self) -> Option<N>
```
would still be better than calling `.unwrap()` inside this function

lazovicff: This could be a solution until we implement a proper error handling.
But also I see that `quotient.add()` is never called, so it might not be needed

4rmanc: Okay. Also, why coverage check is failing? It seems like a problem from server side.

4rmanc (approved):

PR #112, opened 2022-11-08
Integer Chip

lazovicff:


lazovicff:
We should have 1 more test case:
- Using the production flow for proving/verifying - One test case that does both addition and multiplication

Also, TestCircuit should accept an array of Integers, and inside the synthesis, we accumulate them. In case of 2 numbers, pass an array with 2 integers.

4rmanc (approved):

PR #117, opened 2022-11-25
Scalar Multiplication fix

4rmanc:
@lazovicff The problem in this scalar multiplication is that the `add` function causes problem when we try to do a operation with value `zero`. [In here](https://github.com/eigen-trust/eigen-trust/blob/ecc-fix/circuit/src/ecc/native.rs#L140-L152), I added a flag for the manual assign operation for the first value and the algorithm works. Should we add this edge cases to inside `add` and `double` function? In the projective space, point (0, 1, 0) is the [point at the infinity](https://crypto.stanford.edu/pbc/notes/elliptic/group.html). When we take this value to the `affine representation` it becomes (0, 1), so that is why our `zero` value breaks this algorithms.

### DIFF
--- a/circuit/src/ecc/native.rs
+++ b/circuit/src/ecc/native.rs
@@ -17,9 +17,7 @@ use crate::integer::{
 	native::{Integer, ReductionWitness},
 	rns::RnsParams,
 };
-use halo2wrong::halo2::arithmetic::{Field, FieldExt};
-use num_bigint::BigUint;
-use num_traits::{FromPrimitive, Zero};
+use halo2wrong::halo2::arithmetic::FieldExt;
 
 /// Structure for the EcPoint
 #[derive(Clone, Debug)]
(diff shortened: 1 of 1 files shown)


circuit/src/ecc/native.rs
```diff
@@ -52,9 +50,9 @@ where
 		Self::new(Integer::zero(), Integer::one())
 	}
 
-	/// Create a new object with x = 0 and y = 0
-	pub fn identity() -> Self {
-		Self::new(Integer::zero(), Integer::zero())
+	/// Create a new object with x = 0 and y = 1
+	pub fn one() -> Self {
+		Self::new(Integer::one(), Integer::one())
```
lazovicff, line 55: If we need to keep this function, then the comment should be fixed

lazovicff:
Nice job, I'm ok with the implementation. One small comment before we merge

PR #123, opened 2022-12-12
Merkle Tree

4rmanc:


lazovicff:
Found this Merkle tree illustration on Google, which can serve as a reference:
MT

In my experience, the best structure for merkle tree representation would be a map, where each level holds its on array of nodes. Lets say we have a tree of height 4, which means we have 2 ** 4 = 16 leaf nodes:
```rust
map {
  4 => vec![0]
  3 => vec![0, 1]
  2 => vec![0, 1, 2, 3]
  1 => vec![0, 1, 2, 3, 4, 5, 6, 7]
  0 => vec![0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15]
}
```
Each key in the map is the level index. The leaf level is `0`, first is 1`, ..., root is `4`.

This will make it easier to find the path from particular index to the root.

This is just a suggestion if you are stuck, but we should look for simplest ways to represent this.

circuit/src/merkle_tree/native.rs
```diff
@@ -0,0 +1,158 @@
...
+const WIDTH: usize = 5;
+
+#[derive(Clone, Debug)]
+struct MerkleTree<F: FieldExt, P>
+where
+	P: RoundParams<F, WIDTH>,
+{
+	_params: PhantomData<P>,
+	root: Option<F>,
+	indexes: Vec<Option<F>>,
```
lazovicff, line 16: Not sure why its called `indexes`, when it stores node values.

circuit/src/merkle_tree/native.rs
```diff
@@ -0,0 +1,158 @@
...
+			first_leaf_index *= 2;
+			nodes = 2 * nodes;
+		}
+		first_leaf_index -= 1;
+		nodes *= 2;
+		let indexes = vec![None; nodes - 1];
+		MerkleTree { _params: PhantomData, root: None, indexes, height, first_leaf_index }
+	}
+
+	fn implement_leaves(self: &mut MerkleTree<F, P>, values: Vec<F>) -> &Self {
```
lazovicff, line 39: Not sure the name `implement_leaves` is the best choice. Add leaves sounds more clear

circuit/src/merkle_tree/native.rs
```diff
@@ -0,0 +1,158 @@
...
+		self
+	}
+
+	fn build_tree(self: &mut MerkleTree<F, P>, height: usize) -> &Self {
+		if height > 0 {
+			let mut new_tree = MerkleTree::<F, P>::new(height - 1);
+			let first_leaf_index = (new_tree.first_leaf_index * 2) + 1;
+			let mut hash = Vec::new();
+			for i in 0..first_leaf_index {
+				let mut hasher: PoseidonSponge<F, WIDTH, P> = PoseidonSponge::new();
```
lazovicff, line 52: Not sure why we are using sponge, when we are hashing only 2 values.

circuit/src/merkle_tree/native.rs
```diff
@@ -0,0 +1,158 @@
...
+			for i in new_tree.first_leaf_index..new_tree.indexes.len() {
+				self.indexes[i] = new_tree.indexes[i];
+			}
+			self.build_tree(new_tree.height);
+		}
+		self.root = self.indexes[0];
+		self
+	}
+
+	fn find_path(self: &mut MerkleTree<F, P>, value: Option<F>, index: Option<usize>) -> Vec<F> {
```
lazovicff, line 71: 1) `self: &mut MerkleTree<F, P>` can be just replaced by `&mut self`. Same goes for other functions

2) I guess you wanted to have the option to find path either by value or index. I think having only by value is sufficient for our use-case

circuit/src/merkle_tree/native.rs
```diff
@@ -0,0 +1,158 @@
...
+		let mut is_inside = None;
```
lazovicff, line 72: Name `is_inside` suggests to the reader that the variable stores some kind of boolean, but it stores an index of the leaf

circuit/src/merkle_tree/native.rs
```diff
@@ -0,0 +1,158 @@
...
+		}
+		path_vec.push(self.root.unwrap());
+
+		for i in 0..path_vec.len() - 2 {
+			let mut hasher: PoseidonSponge<F, WIDTH, P> = PoseidonSponge::new();
+			if i % 2 == 0 {
+				hasher.update(&[path_vec[i], path_vec[i + 1]]);
+				assert!(Self::is_inside(&path_vec, hasher.squeeze()));
+			}
+		}
```
lazovicff, line 106: This should not be here. This logic will be used for path verification. So perhaps, a `Path` struct is needed that implements this

circuit/src/merkle_tree/native.rs
```diff
@@ -0,0 +1,158 @@
...
+
+	fn is_inside(vector: &Vec<F>, value: F) -> bool {
+		let mut answer = false;
+		for i in 0..vector.len() {
+			if value == vector[i] {
+				answer = true;
+			}
+		}
+		answer
+	}
```
lazovicff, line 118: This entire function can be replaced by `vec.contains(value)`

circuit/src/merkle_tree/native.rs
```diff
@@ -0,0 +1,158 @@
...
+
+	use super::MerkleTree;
+
+	#[test]
+	fn should_build_tree() {
+		let mut merkle = MerkleTree::<Fr, Params>::new(2);
+		merkle
+			.implement_leaves([(); 4].map(|_| <Fr as Field>::random(rand::thread_rng())).to_vec());
+
+		merkle.build_tree(merkle.height);
```
lazovicff, line 135: This function accepts a value that is a part of `merkle` which is the object a function is part of. So this argument is not needed.

circuit/src/merkle_tree/native.rs
4rmanc: I did use this because Poseidon accepts minimum width of 5. I thought I can construct an array with width 5 but we will only use 2 for each loop. Is this better use case then the sponge?

circuit/src/merkle_tree/native.rs
4rmanc: I did return Path { path_arr, value} for this function

lazovicff: Sure, now we need a function inside `Path`, called `verify` or similar, that includes the logic highlighted above.

circuit/src/merkle_tree/native.rs
lazovicff: I think we should use sponge for arbitrary length arrays where length is > WIDTH. In cases like this, when we are sure about our input width, we could use normal poseidon and add padding manually, e.g.:
`[value1, value2, zero, zero, zero]`

circuit/src/merkle_tree/native.rs
4rmanc: Okay, I got the point. Thank you

circuit/src/merkle_tree/native.rs
```diff
@@ -0,0 +1,166 @@
...
+					let hasher: Poseidon<F, WIDTH, P> = Poseidon::new([
+						self.nodes[first_leaf_index + i].unwrap(),
+						self.nodes[first_leaf_index + i + 1].unwrap(),
+						F::zero(),
+						F::zero(),
+						F::zero(),
+					]);
+					let hashes = hasher.permute();
+
+					hash.push(hashes[0] + hashes[1]);
```
lazovicff, line 71: Just curious, why are we adding hash[0] and hash[1]?

4rmanc: I wanted to concatenate both hash values by adding them.

circuit/src/merkle_tree/native.rs
```diff
@@ -0,0 +1,173 @@
...
+	P: RoundParams<F, WIDTH>,
+{
+	/// Create a new empty Merkle Tree with given height
+	fn new(height: usize) -> Self {
+		// 0th level is the leaf level and the max level is the root level
+		let mut nodes = HashMap::new();
+		for i in 0..height + 1 {
+			let num_nodes = pow(2, height - i);
+			let nodes_vec = vec![F::zero(); num_nodes];
+			nodes.insert(i, nodes_vec);
```
lazovicff, line 35: While leaf nodes can be zero, it doesnt make any sense for upper level nodes to be zero, they should always be hashes of 2 child nodes.

I think we should merge all three functions into one, such that:
You pass `height` as the height of the tree, and `leaves` as leaves of the tree, but the length of leaves array should be <= 2 ** height. If it's less than that, the rest of the leaves should be defaulted to zero, so that the leaves array is filled to the next power of two.

4rmanc: I also took out height variable from function inputs. I did height calculations inside the circuit so we don't have to concern about if number of leaves is bigger than the 2**height.

circuit/src/merkle_tree/native.rs
```diff
@@ -0,0 +1,183 @@
...
+		for i in 0..height {
+			if height == 1 {
+				height = i;
+				if pow(2, i) < leaves.len() {
+					height += 1;
+				}
+				break;
+			}
+			height /= 2;
+		}
```
lazovicff, line 43: Next power of 2 for leaves length can be calculated by simply calling `next_power_of_two` function on `usize`.
`height` is then `log2` of that calculated value

circuit/src/merkle_tree/native.rs
lazovicff: I think in our use-case we want to explicitly pass the `height`, since it will be predetermined inside a smart contract or similar. In case of leaves array being bigger than height, we can panic, by adding `assert!(leaves.len() <= 2**height)` or similar

circuit/src/merkle_tree/native.rs
lazovicff: But calling `next_power_of_two` wont be necessary I think, given this: https://github.com/eigen-trust/eigen-trust/pull/123#discussion_r1054276066

4rmanc: Okay, thank you for the information

circuit/src/merkle_tree/native.rs
```diff
@@ -0,0 +1,172 @@
...
+
+	/// Find path for the given value to the root
+	fn find_path(&mut self, value: F) -> Path<F, P> {
+		let mut value_index = None;
+		for i in 0..self.nodes[&0].len() {
+			if value == self.nodes[&0][i] {
+				value_index = Some(i);
+				break;
+			}
+		}
```
lazovicff, line 66: To simplify this search, you can use:
```rust
let value_index = self.nodes[&0].iter().position(|x| x == &value);
```

circuit/src/merkle_tree/native.rs
```diff
@@ -0,0 +1,172 @@
...
+const WIDTH: usize = 5;
+
+#[derive(Clone, Debug)]
+/// MerkleTree structure
+struct MerkleTree<F: FieldExt, P>
+where
+	P: RoundParams<F, WIDTH>,
+{
+	/// PhantomData for the params
+	_params: PhantomData<P>,
```
lazovicff, line 17: PhantomData usually goes at the bottom of the struct, if we wanted to follow the convention from other structs.

circuit/src/merkle_tree/native.rs
```diff
@@ -0,0 +1,172 @@
...
+}
+
+#[derive(Clone)]
+/// Path structure
+struct Path<F: FieldExt, P>
+where
+	P: RoundParams<F, WIDTH>,
+{
+	/// PhantomData for the params
+	_params: PhantomData<P>,
```
lazovicff, line 93: Same as https://github.com/eigen-trust/eigen-trust/pull/123/files#r1054321612

circuit/src/merkle_tree/native.rs
```diff
@@ -0,0 +1,172 @@
...
+			for i in 0..nodes[&level].len() {
+				if i % 2 == 0 {
+					let hasher: Poseidon<F, WIDTH, P> = Poseidon::new([
+						nodes[&level][i],
+						nodes[&level][i + 1],
+						F::zero(),
+						F::zero(),
+						F::zero(),
+					]);
+					hash.push(hasher.permute()[0]);
```
lazovicff, line 50: We could do a little change here to avoid deep nesting and make it more visually clear.
1) we can replace `i % 2 == 0` with `i % 2 != 0 { continue; }`
2) Define Poseidon inputs before calling the `new` function:
```rust
let pos_inputs = [
  nodes[&level][i],
  nodes[&level][i + 1],
  F::zero(),
  F::zero(),
  F::zero(),
]
let hasher: Poseidon<F, WIDTH, P> = Poseidon::new(pos_inputs);
```

circuit/src/merkle_tree/native.rs
```diff
@@ -0,0 +1,172 @@
...
+		for i in 0..self.path_vec.len() - 1 {
+			if i % 2 == 0 {
+				let hasher: Poseidon<F, WIDTH, P> = Poseidon::new([
+					self.path_vec[i],
+					self.path_vec[i + 1],
+					F::zero(),
+					F::zero(),
+					F::zero(),
+				]);
+				is_satisfied = is_satisfied | self.path_vec.contains(&(hasher.permute()[0]));
```
lazovicff, line 116: Same as https://github.com/eigen-trust/eigen-trust/pull/123/files#r1054321755

circuit/src/merkle_tree/native.rs
lazovicff: I should add that this approach of finding leaf nodes by values will break in case we have leaves with the same value, but we can leave that concern for the future

circuit/src/merkle_tree/native.rs
```diff
@@ -0,0 +1,172 @@
...
+		// 0th level is the leaf level and the max level is the root level
+		let mut nodes = HashMap::new();
+		// Assign zero to the leaf values if they are empty
+		for i in leaves.len()..pow(2, height) {
+			leaves.push(F::zero());
+		}
+		nodes.insert(0, leaves);
+
+		for level in 0..height {
+			let mut hash = Vec::new();
```
lazovicff, line 40: This should be plural: `hashes`

lazovicff (approved):

circuit/src/merkle_tree/mod.rs
```diff
@@ -0,0 +1,310 @@
...
+					[path_vec[i + 2], F::zero()],
+					hashes[0].clone(),
+				));
+				root = Some(hashes[0].clone());
+			} else {
+				set = Some(FixedSetChip::<F, 2>::new(
+					[path_vec[i + 2], path_vec[i + 3]],
+					hashes[0].clone(),
+				));
+			}
```
lazovicff, line 96: I think this path structure is forcing us to write a lot of complicated code. How about changing the struct to accept 2 generic params:
```rust
struct Path<..., const HEIGHT: usize, const ARITY: usize> {
  nodes: [[F; ARITY]; HEIGHT];
}
```

This will force use to make a correct structure before passing it into the chip

4rmanc: @lazovicff Is there any significant difference between using an array or a vector in the project? I know that array is more stable and kept at stack of the memory, but is that crucial that much? Because if we use `Vec<Vec<F>>` we won't need `ARITY` and `HEIGHT` constants.

lazovicff: We are using arrays because they are more deterministic. We know exactly what to expect and that simplifies the code.
Why do you think its bad to have const generics like ARITY and HEIGHT?

lazovicff: If having more const generics are confusing to work with, you can use hardcoded arity of 2

4rmanc: No, I am actually comfortable with using const generics, but I thought it will be harder at the readability side. I will focus on using arrays.

circuit/src/merkle_tree/native.rs
```diff
@@ -0,0 +1,155 @@
...
+		// algorithm.
+		for level in 0..merkle_tree.height {
+			if value_index % 2 == 1 {
+				path_arr[level][0] = merkle_tree.nodes[&level][value_index - 1];
+				path_arr[level][1] = merkle_tree.nodes[&level][value_index];
+			} else {
+				path_arr[level][0] = merkle_tree.nodes[&level][value_index];
+				path_arr[level][1] = merkle_tree.nodes[&level][value_index + 1]
+			}
+			value_index = value_index / 2;
```
lazovicff, line 95: `ARITY` stands for the number of nodes to be hashed together at each level. In this part we are ignoring the `ARITY` constant and treating this tree as binary (2 nodes at each level).
You can either remove ARITY constant and replace it with 2, or handle the logic.

4rmanc: I was trying to do some logic to keep last variable (root) as a single element and I miss to change that back to 2. Fixing it immediately and doing last checks for the comment lines.

circuit/src/merkle_tree/mod.rs
```diff
@@ -0,0 +1,297 @@
...
+		let temp = meta.advice_column();
+		let set = FixedSetChip::<F, 2>::configure(meta);
+		let poseidon = PoseidonChip::<_, WIDTH, P>::configure(meta);
+		meta.enable_equality(temp);
+		MerklePathConfig { temp, set, poseidon }
+	}
+
+	/// Synthesize the circuit.
+	pub fn synthesize(
+		&self, config: MerklePathConfig, path_arr: [[F; 2]; LENGTH], zero: AssignedCell<F, F>,
```
lazovicff, line 62: Not sure why `path_arr` is needed. Its sufficient to check if hash of nodes at level `i` is one of the nodes at level `i + 1`. So we could use just `self.nodes` for this

lazovicff: Ah, ok, `FixedSetChip` is not written properly. I will fix that at some point

lazovicff (approved):

PR #130, opened 2023-01-10
API work (Reduction Witness)

4rmanc:


circuit/src/integer/mod.rs
```diff
@@ -250,100 +91,611 @@ where
...
+	_rns: PhantomData<P>,
+}
+
+impl<W: FieldExt, N: FieldExt, const NUM_LIMBS: usize, const NUM_BITS: usize, P>
+	IntegerReduceChip<W, N, NUM_LIMBS, NUM_BITS, P>
+where
+	P: RnsParams<W, N, NUM_LIMBS, NUM_BITS>,
+{
+	/// Creates a new reduce chip
+	pub fn new(assigned_integer: &AssignedInteger<W, N, NUM_LIMBS, NUM_BITS, P>) -> Self {
```
lazovicff, line 132: Its is a convention to pass values in the constructor instead of references.
Apply in all similar cases

circuit/src/integer/mod.rs
```diff
@@ -250,100 +91,611 @@ where
...
+			let reduce_q_exp = v_cells.query_advice(common.advice[NUM_LIMBS], Rotation::prev());
+			let mut limbs_exp = [(); NUM_LIMBS].map(|_| None);
+			let mut residues_exp = Vec::new();
+			let mut result_exp = [(); NUM_LIMBS].map(|_| None);
+			for i in 0..NUM_LIMBS {
+				t_exp[i] = Some(v_cells.query_advice(common.advice[i], Rotation::prev()));
+				limbs_exp[i] = Some(v_cells.query_advice(common.advice[i], Rotation::cur()));
+				if i < NUM_LIMBS / 2 {
+					residues_exp.push(v_cells.query_advice(common.advice[i], Rotation::next()));
+				}
```
lazovicff, line 166: Can we move this out of this loop and put it inside its own loop, just to avoid nesting and make it more clear.
Apply in all similar cases

circuit/src/integer/mod.rs
```diff
@@ -250,100 +91,611 @@ where
...
+	fn synthesize(
+		self, common: &CommonConfig, selector: &Selector, mut layouter: impl Layouter<N>,
+	) -> Result<Self::Output, Error> {
+		let reduction_witness = self.assigned_integer.integer.reduce();
 		layouter.assign_region(
 			|| "reduce_operation",
 			|mut region: Region<'_, N>| {
-				config.reduce_selector.enable(&mut region, 0)?;
-				Self::assign(&x_limbs, None, &rw, &config, &mut region, 0)
+				selector.enable(&mut region, 1)?;
```
lazovicff, line 195: Can we use RegionCtx here to enable the selector, and then pass it to `IntegerAssign::assign`.
Apply in all similar cases

circuit/src/ecc/mod.rs
```diff
@@ -2,628 +2,567 @@
...
-		Ok((x, y))
+		// numerator = other.y.sub(&self.y);
+		let numerator_chip = IntegerSubChip::new(&q_y_reduced, &p_y_reduced);
+		let numerator = numerator_chip
+			.synthesize(
+				&common,
+				&config.integer_sub_selector,
+				layouter.namespace(|| "numerator"),
+			)
+			.unwrap();
```
lazovicff, line 145: We should propagate the error down the stack using `?`.
Apply to similar cases

lazovicff (approved):

lazovicff:
Approved, but in the next PR we should focus more on making the code more readable and structuarly and contextually correct. e.g. https://github.com/eigen-trust/protocol/blob/ecc-fix/circuit/src/integer/mod.rs#L128-L155:
- Should put contextually similar code blocks together, use empty lines to separate those blocks
- When querying rows, we should do it in sequential order - cur(), next(), 2, etc.

lazovicff (approved):

PR #146, opened 2023-01-19
Ladder implementation

4rmanc:


4rmanc:
Note that if any one of the `x` coordinates from `p` and `q`  is [equal to another coordinate variable](https://github.com/eigen-trust/protocol/blob/ladder-implementation/circuit/src/ecc/mod.rs#L1073-L1074) from both, it will give an addition constraint error. 
It is also noted in the [paper](https://hackmd.io/ncuKqRXzR-Cw-Au2fGzsMg?view) that we need to satisfy conditions below:
* `P_0 != 0` and `P_1 != 0` to skip infinity checks
* `P_0 != P_1` to just constain addition
* `P_0 != -P_1` to avoid result goes to infinity.

Will do further research about this topic.

lazovicff (approved):

Issue #126, opened 2022-12-16
Rescue Prime Chip


lazovicff:
The native version of the chip is [already developed](https://github.com/eigen-trust/eigen-trust/tree/master/circuit/src/rescue_prime). We should make a Halo2 chip version, modelled after our [Poseidon Chip](https://github.com/eigen-trust/eigen-trust/tree/master/circuit/src/poseidon).

It should implement Chip trait, since it only needs one selector.
One thing to note is that the maximum width of the permutation state could be `5`, which is limited by the amount of fixed columns we have in our CommonConfig/MainConfig. for each permutation of `width`, the amount of fixed columns required is `width * 2`, and we have `10` fixed columns available.

EDIT: Make a RescuePrimeSpongeChipset that includes already built [SpongeChip](https://github.com/eigen-trust/protocol/blob/master/circuit/src/poseidon/sponge.rs#L15-L86). It should be modelled after [PoseidonSpongeChipset implementation](https://github.com/eigen-trust/protocol/blob/master/circuit/src/poseidon/sponge.rs#L128-L180).
Native version of the RescuePrime sponge should be implemented as well.

lazovicff:
Pull requests for this issue:
https://github.com/eigen-trust/protocol/pull/148
https://github.com/eigen-trust/protocol/pull/152

A bonus research task was also done in this milestone, related to #141:
https://github.com/eigen-trust/protocol/pull/154

PR #148, opened 2023-01-26
Implement the `RescuePrimeChip` in halo2

duguorong009:
Reference: #126


circuit/src/rescue_prime/mod.rs
```diff
@@ -1,2 +1,334 @@
...
+			//   |  3  |       | sbox_invs(2) |               |            |
+			//   |  4  |   3   |  state(3)    |    const(3)   |     *      |
+			//                 ...
+
+			let mut exprs = vec![];
+			let s_cells = v_cells.query_selector(selector);
+
+			let mut state = state_columns.map(|c| v_cells.query_advice(c, Rotation::cur()));
+			let round_constants = rc_columns.map(|c| v_cells.query_fixed(c, Rotation::cur()));
+			let next_round_constants = rc_columns.map(|c| v_cells.query_fixed(c, Rotation(2)));
```
lazovicff, line 87: Why are we using `Rotation(2)` instead of `Rotation::next()`?

lazovicff: Ah ok, I got it

lazovicff (dismissed):

lazovicff (approved):

PR #154, opened 2023-02-06
Dynamic Sets

lazovicff:


circuit/src/native.rs
```diff
@@ -0,0 +1,403 @@
...
+		let op = Opinion::new(sig, message_hashes[0], scores);
+
+		set.update_op(pk3, op);
+
+		set.converge();
+	}
+
+	#[test]
+	#[should_panic]
+	fn test_add_three_members_with_two_opinions() {
```
lazovicff, line 360: We should have two versions of this test
1) We should have 2 peers giving scores to each others, and the third giving scores to invalid peers, so the third one should be invalidate and the scores should successfully converge
2) Peer 1 should give score to Peer 2, Peer 2 should give score to Peer 3, and Peer 3 should give scores to invalid peers. Peer 3 will get invalidated, which means Peer 2 will get invalidated, which also means Peer 1 will get invalidated. The scores should not be able to converge

circuit/src/native.rs
```diff
@@ -0,0 +1,403 @@
...
+			} else {
+				let mut ops_i = self.ops.get_mut(&pk).unwrap();
+
+				// Nullify scores that are given to wrong member at specific index
+				let mut score_sum = Fr::zero();
+				for j in 0..NUM_NEIGHBOURS {
+					let (pk_j, score) = ops_i.scores[j].clone();
+					let true_score =
+						if j != i && self.set[j].0 == pk_j { score } else { Fr::zero() };
+					ops_i.scores[j].1 = true_score;
```
lazovicff, line 81: Instead of mutating the original struct, we should make a copy `filtered_scores` and modify that matrix instead. Same goes for the `filtered_set`

circuit/src/native.rs
```diff
@@ -0,0 +1,403 @@
...
+					// Normalize the scores
+					// Distribute the credits(INITIAL_SCORE) to the valid opinion values
+					// We assume that the initial credits of peer is constant(INITIAL_SCORE)
+					for j in 0..NUM_NEIGHBOURS {
+						let (_, op_score) = ops_i.scores[j].clone();
+						ops_i.scores[j].1 = op_score * score_sum.invert().unwrap() * initial_score;
+					}
+				}
+			}
+		}
```
lazovicff, line 100: Can we make this more readable by making less logic branches?
Perhaps break into several loops instead of having just one.

duguorong009: Yes, I done some refactoring in https://github.com/eigen-trust/protocol/pull/154/commits/0c365b1ffd58962ddf16563fc280a1391b3aca91

circuit/src/native.rs
duguorong009: Added the `filtered` variables in https://github.com/eigen-trust/protocol/pull/154/commits/0c365b1ffd58962ddf16563fc280a1391b3aca91

circuit/src/native.rs
duguorong009: I think the first test is done in `test_add_three_members_with_two_opinions()` test example.

For the second test, I suspect if it needs more complicated algorithm(related to graph?).
It seems that I should check the connectivity between peers, and it needs much more than forward/backward loop.

circuit/src/native.rs
```diff
@@ -0,0 +1,561 @@
...
+		let initial_score = Fr::from_u128(INITIAL_SCORE);
+		let index = pos_from.unwrap();
+		self.set[index] = (from, initial_score);
+
+		self.ops.insert(from, op);
+	}
+
+	pub fn converge(&mut self) -> [Fr; NUM_NEIGHBOURS] {
+		let mut filtered_set = self.set.clone();
+		let mut filtered_ops: HashMap<PublicKey, Opinion> = HashMap::new();
```
lazovicff, line 84: The goal is to avoid mutating the original variables, so here we should do `self.ops.clone()`

circuit/src/native.rs
```diff
@@ -0,0 +1,652 @@
...
+					} else if pk_j == pk_i {
+						PublicKey::default()
+					} else {
+						pk_j
+					};
+					let true_score = if true_pk == pk_i || true_pk == PublicKey::default() {
+						Fr::zero()
+					} else {
+						score
+					};
```
lazovicff, line 206: Can we do 2 checks:
1) if pk_i == pk_j
2) if p_i == null

If any of these is true, replace the score with 0

circuit/src/native.rs
```diff
@@ -0,0 +1,652 @@
...
+				// Example 2:
+				// 	filtered_set => [p1, p2, p3]
+				//	Peer1 opinion
+				// 		[(p1, 0), (p3, 10), (null, 0)]
+				//   => [(p1, 0), (p2, 0),  (p3, 0)]
+				let op_score_sum =
+					ops_i.scores.iter().fold(Fr::zero(), |acc, &(_, score)| acc + score);
+				if op_score_sum == Fr::zero() {
+					panic!("The peer {i} didn't give any score to other peers!");
+				}
```
lazovicff, line 227: We shouldn't panic here. Instead we should distribute the scores to everyone equally

duguorong009: Hmm, the example case of this scenario is like:

`Set = [p1, p2, p3]`
Peer1 opinion(op1) =` [(p1, 10), (p2, 0), (p3, 0)]`
Then, the op1 becomes` [(p1, 0), (p2, 0), (p3, 0)] `after first filtering.
It means that the `p1` didn't give any score to `p2` and `p3`.

Do you think we should still give the scores even when `p1` originally does not give score?

circuit/src/native.rs
duguorong009: Yes, I think I over-complicated the logic here.
Please check update in https://github.com/eigen-trust/protocol/pull/154/commits/3e727c695a242bbd58c5b73102c3307a95f5a8eb

circuit/src/native.rs
lazovicff: Yes, I think we should

duguorong009: Ok, here is update commit (https://github.com/eigen-trust/protocol/pull/154/commits/9850d1b72e707d73162413c95512ad37feaa5714)

lazovicff:
> @lazovicff
I have one interesting case.
>
> Let us assume:
Set = [p1, p2, p3]
Peer1 opinion op1 = [(p1, 0), (p3, 10), (p2, 40)]
>
> According to current algorithm, this op1 becomes [(p1, 0), (p2, 1), (p3, 1)] after filtering.
The reason is that the order of opinion scores matters, and the score is equally distributed.
I think this is not fair, since the peer1 gives his opinion correctly, just in wrong order.
>
> What do you think about this?

Yes, I'm aware of this. For now, we wont handle this case. But, in the future we would have to run some kind of sorting algorithm inside the circuit,

lazovicff:
I see that you simplified the algorithm a lot, good job!
We could make it more readable by reducing nesting.

circuit/src/native.rs
```diff
@@ -0,0 +1,628 @@
...
+		let pk1 = sk1.public();
+
+		set.add_member(pk1);
+
+		// Re-adding the member should panic
+		set.add_member(pk1);
+	}
+
+	#[test]
+	#[should_panic]
```
lazovicff, line 289: I this we could also avoid panic in this scenario, by:

If member is in the set but didn't add his opinion, we should assign default one, e.g.:
if `set = [peer1, peer2, peer3]`
if `peer1` didn't add his opinion, we assign `peer1_op = [(peer1, 0), (peer2, 1), (peer3, 1)]`

duguorong009:
> This algorithm should panic only in case we have less than 2 peers in the set, regardless if they added their opinion or not

Okay, so the point is that we should use the default opinion when the peer did NOT add/sign the opinion.
Please check https://github.com/eigen-trust/protocol/pull/154/commits/5c125b705cabca455881c0d1368bb82082786573 for implementation.

duguorong009 (approved):

PR #155, opened 2023-02-06
mul_scalar fix

4rmanc:


circuit/src/ecc/native.rs
```diff
@@ -50,6 +56,41 @@ where
...
+		let x = BigUint::from_str(
+			"13309346756614635333825509300060708409153440696756098989763444826124962874203",
+		)
+		.unwrap();
+		let y = BigUint::from_str(
+			"19567858517750281949071914889393977810091135383309479881360435638939337926797",
+		)
+		.unwrap();
+		Self::new(Integer::new(x), Integer::new(y))
+	}
```
lazovicff, line 92: To avoid parsing these numbers and decomposing them every time, we should have hardcoded limbs here.
Also we should consider moving this to Rns.

circuit/src/ecc/native.rs
```diff
@@ -125,38 +166,34 @@ where
...
-				} else {
-					r = r.add(&exp.clone());
-				}
-			}
-			exp = exp.double();
+		let r_init = Self::to_add();
+		let exp: EcPoint<W, N, NUM_LIMBS, NUM_BITS, P> = self.clone();
+
+		// Converts given input to its bit by Scalar Field's bit size
+		let mut bits = to_bits::<254>(le_bytes);
```
lazovicff, line 175: Can we use N::NUM_BUTS instead of hardcoding 254

circuit/src/ecc/native.rs
```diff
@@ -125,38 +166,34 @@ where
...
+		let table = [r_init.clone(), exp.clone().add(&r_init)];
+		let mut acc = Self::select(bits[0], table.clone());
+
+		// To avoid P_0 == P_1
+		acc = acc.double();
+		if bits[1] {
+			acc = acc.add(&table[1]);
+		} else {
+			acc = acc.add(&table[0]);
+		}
```
lazovicff, line 187: We should use `Self::select` here, to stay consistent

circuit/src/ecc/native.rs
lazovicff: Ok, you can leave that one out. We have to find solution to mul_scalar's than passing bytes. I created a task for this: https://github.com/eigen-trust/protocol/issues/158

lazovicff (approved):

PR #165, opened 2023-02-24
Solidity plonk verifier

lazovicff:


circuit/src/utils.rs
```diff
@@ -26,7 +26,104 @@ use halo2::{
...
+	let mut buffer = String::new();
+
+	// Read file into string.
+	reader.read_to_string(&mut buffer).unwrap();
+
+	buffer
+}
+
+/// Reads raw bytes from the file
+pub fn read_bytes_data(name: &str) -> Vec<u8> {
```
duguorong009, line 64: How about `bin_file_name` or `file_name`, instead of `name`?

circuit/src/utils.rs
```diff
@@ -26,7 +26,104 @@ use halo2::{
...
+	read_bytes(bin_path)
+}
+
+/// Write bytes to a file
+pub fn write_bytes(bytes: Vec<u8>, path: impl AsRef<Path>) -> Result<(), IoError> {
+	write(path, bytes)
+}
+
+/// Write bytes to data directory
+pub fn write_bytes_data(bytes: Vec<u8>, name: &str) -> Result<(), IoError> {
```
duguorong009, line 76: Same as above.
I think it is better to use clear variable name, like `file_name`, instead of just `name`.

duguorong009 (commented):
I left some comments.
Simply, I think it is better to give clear variable names to utility functions in `circuit/src/utils.rs`.

circuit/src/utils.rs
lazovicff: Sure, we can do that

PR #168, opened 2023-03-06
Aggregator

4rmanc:


circuit/src/verifier/loader/mod.rs
```diff
@@ -588,6 +588,7 @@ where
 			&Self::LoadedEcPoint,
 		)],
 	) -> Self::LoadedEcPoint {
+		// TODO: Define config and aux inside the iteratior
```
4rmanc: @lazovicff Why did we carry config definition inside the iterator? It will be defined twice with this way, in the other hand if it is outside it will be defined once. Is it because of the readability?

lazovicff: I think it will make more sense because `pairs` can be empty, in which case it will panic

lazovicff (approved):

PR #169, opened 2023-03-06
Implement the `LookupRangeCheckChipset` for better range check functionality

duguorong009:
Reference: #127

### DIFF
--- a/circuit/Cargo.toml
+++ b/circuit/Cargo.toml
@@ -9,7 +9,7 @@ edition = "2021"
 plotters = { version = "0.3.0" }
 rand = "0.8"
 hex = { version = "0.4", default-features = false, features = ["alloc"] }
-halo2 = { package = "halo2_proofs", git = "https://github.com/privacy-scaling-explorations/halo2", tag = "v2023_02_02" }
+halo2 = { package = "halo2_proofs", git = "https://github.com/privacy-scaling-explorations/halo2", tag = "v2023_02_02", features = ["dev-graph"] }
 blake = "2.0.2"
 itertools = "0.10.3"
 num-bigint = {version = "0.2.2", features = ["rand"]}
--- a/circuit/src/gadgets/mod.rs
+++ b/circuit/src/gadgets/mod.rs
@@ -8,3 +8,6 @@ pub mod lt_eq;
 pub mod main;
 /// Set membership gadget
 pub mod set;
+
+/// Range check gadget
+pub mod range;
--- a/circuit/src/gadgets/range.rs
+++ b/circuit/src/gadgets/range.rs
@@ -80,8 +80,8 @@ impl<F: FieldExt + PrimeFieldBits, const K: usize, const S: usize> Chip<F>
 				// Assign shifted value
 				ctx.next();
 
-				let shiftted_word = self.x.value().into_field() * F::from(1 << (K - S));
-				ctx.assign_advice(common.advice[0], shiftted_word.evaluate())?;
+				let shifted_word = self.x.value().cloned() * Value::known(F::from(1 << (K - S)));
+				ctx.assign_advice(common.advice[0], shifted_word)?;
 				ctx.enable(selector.clone())?;
 
 				// Assign 2^{-S} from a fixed column.
@@ -152,6 +152,7 @@ impl<F: FieldExt + PrimeFieldBits, const K: usize, const N: usize> Chip<F>
 			|| "range check",
 			|region| {
 				let mut ctx = RegionCtx::new(region, 0);
+				let x = ctx.copy_assign(common.advice[0], self.x.clone())?;
 
 				let num_bits = K * N;
 
@@ -170,7 +171,7 @@ impl<F: FieldExt + PrimeFieldBits, const K: usize, const N: usize> Chip<F>
 					.transpose_vec(N)
 				};
 
-				let mut zs = vec![self.x.clone()];
+				let mut zs = vec![x.clone()];
 
 				// Assign cumulative sum such that
 				//          z_i = 2^{K}⋅z_{i + 1} + a_i
@@ -178,7 +179,7 @@ impl<F: FieldExt + PrimeFieldBits, const K: usize, const N: usize> Chip<F>
 				//
 				// For `element` = a_0 + 2^10 a_1 + ... + 2^{120} a_{12}}, initialize z_0 =
 				// `element`. If `element` fits in 130 bits, we end up with z_{13} = 0.
-				let mut z = self.x.clone();
+				let mut z = x.clone();
 				let inv_two_pow_k = F::from(1u64 << K).invert().unwrap();
 				for word in words {
 					// Enable q_lookup on this row
@@ -190,12 +191,11 @@ impl<F: FieldExt + PrimeFieldBits, const K: usize, const N: usize> Chip<F>
 							z.value().zip(word).map(|(z, word)| (*z - word) * inv_two_pow_k);
 
 						// Assign z_next
+						ctx.next();
 						ctx.assign_advice(common.advice[0], z_val)?
 					};
 
 					zs.push(z.clone());
-
-					ctx.next();
 				}
 
 				ctx.constrain_to_constant(zs.last().unwrap().clone(), F::zero())?;
@@ -251,6 +251,10 @@ mod tests {
 		plonk::{Circuit, ConstraintSystem, Error},
 	};
 
+	const K: usize = 8;
+	const S: usize = 3;
+	const N: usize = 2;
+
 	#[derive(Debug, Clone)]
 	enum Gadget {
 		ShortWordCheck,
@@ -286,8 +290,8 @@ mod tests {
 
 		fn configure(meta: &mut ConstraintSystem<Fr>) -> TestConfig {
 			let common = CommonConfig::new(meta);
-			let q_bitshift = LookupShortWordCheckChip::<Fr, 8, 3>::configure(&common, meta);
-			let q_running_sum = LookupRangeCheckChip::<Fr, 8, 16>::configure(&common, meta);
+			let q_bitshift = LookupShortWordCheckChip::<Fr, K, S>::configure(&common, meta);
+			let q_running_sum = LookupRangeCheckChip::<Fr, K, N>::configure(&common, meta);
 
 			TestConfig { common, q_bitshift, q_running_sum }
 		}
@@ -300,7 +304,7 @@ mod tests {
 				|| "table_column",
 				|mut table| {
 					// We generate the row values lazily (we only need them during keygen).
-					for index in 0..(1 << 8) {
+					for index in 0..(1 << K) {
 						table.assign_cell(
 							|| "table_column",
 							config.common.table,
@@ -324,7 +328,7 @@ mod tests {
 
 			match self.gadget {
 				Gadget::ShortWordCheck => {
-					let short_word_check_chip = LookupShortWordCheckChip::<Fr, 8, 3>::new(x);
+					let short_word_check_chip = LookupShortWordCheckChip::<Fr, K, S>::new(x);
 					let _ = short_word_check_chip.synthesize(
 						&config.common,
 						&config.q_bitshift,
@@ -332,7 +336,7 @@ mod tests {
 					)?;
 				},
 				Gadget::RangeCheck => {
-					let range_check_chip = LookupRangeCheckChip::<Fr, 8, 16>::new(x);
+					let range_check_chip = LookupRangeCheckChip::<Fr, K, N>::new(x);
 					let _ = range_check_chip.synthesize(
 						&config.common,
 						&config.q_running_sum,
@@ -355,37 +359,39 @@ mod tests {
 		let k = 9;
 		let pub_ins = vec![];
 		let prover = MockProver::run(k, &test_chip, vec![pub_ins]).unwrap();
-		prover.assert_satisfied();
+		assert!(prover.verify().is_ok());
 
 		// Should fail since x is 4 bits
-		let x = Fr::from(0b1000);
+		let x = Fr::from(0b1001);
 
 		let test_chip = TestCircuit::new(x, Gadget::ShortWordCheck);
 
 		let k = 9;
 		let pub_ins = vec![];
-		let _err = MockProver::run(k, &test_chip, vec![pub_ins]).unwrap_err();
+		let prover = MockProver::run(k, &test_chip, vec![pub_ins]).unwrap();
+		assert!(prover.verify().is_err());
 	}
 
 	#[test]
 	fn test_range_check() {
 		// Testing x is 16 bits
-		let x = Fr::from(0b1111111111111111);
+		let x = Fr::from(0b1111111111111100);
 
 		let test_chip = TestCircuit::new(x, Gadget::RangeCheck);
 
 		let k = 9;
 		let pub_ins = vec![];
 		let prover = MockProver::run(k, &test_chip, vec![pub_ins]).unwrap();
-		prover.assert_satisfied();
+		assert!(prover.verify().is_ok());
 
-		// Should fail since x is 17 bits
-		let x = Fr::from(0b1111111111111111000);
+		// Should fail since x is 15 bits
+		let x = Fr::from(0b11111111111111111);
 
 		let test_chip = TestCircuit::new(x, Gadget::RangeCheck);
 
 		let k = 9;
 		let pub_ins = vec![];
-		let _err = MockProver::run(k, &test_chip, vec![pub_ins]).unwrap_err();
+		let prover = MockProver::run(k, &test_chip, vec![pub_ins]).unwrap();
+		assert!(prover.verify().is_err());
 	}
 }
--- a/circuit/src/gadgets/range.rs
+++ b/circuit/src/gadgets/range.rs
@@ -1,0 +1,166 @@
+use std::marker::PhantomData;
+
+use halo2::{
+	circuit::{AssignedCell, Layouter, Value},
+	halo2curves::{group::ff::PrimeFieldBits, FieldExt},
+	plonk::{Constraints, Expression, Selector},
+	poly::Rotation,
+};
+
+use crate::{Chip, Chipset, RegionCtx};
+
+/// Lookup short word check chip
+#[derive(Debug, Clone)]
+pub struct LookupShortWordCheckChip<F: FieldExt + PrimeFieldBits, const K: usize, const S: usize> {
+	x: AssignedCell<F, F>,
+}
+
+impl<F: FieldExt + PrimeFieldBits, const K: usize, const S: usize> Chip<F>
+	for LookupShortWordCheckChip<F, K, S>
+{
+	type Output = AssignedCell<F, F>;
+
+	fn configure(
+		common: &crate::CommonConfig, meta: &mut halo2::plonk::ConstraintSystem<F>,
+	) -> Selector {
+		todo!()
+	}
+
+	fn synthesize(
+		self, common: &crate::CommonConfig, selector: &Selector, layouter: impl Layouter<F>,
+	) -> Result<Self::Output, halo2::plonk::Error> {
+		todo!()
+	}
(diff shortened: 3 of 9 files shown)


circuit/Cargo.toml
```diff
@@ -9,7 +9,7 @@ edition = "2021"
 plotters = { version = "0.3.0" }
 rand = "0.8"
 hex = { version = "0.4", default-features = false, features = ["alloc"] }
-halo2 = { package = "halo2_proofs", git = "https://github.com/privacy-scaling-explorations/halo2", tag = "v2023_02_02" }
+halo2 = { package = "halo2_proofs", git = "https://github.com/privacy-scaling-explorations/halo2", tag = "v2023_02_02", features = ["dev-graph"] }
```
lazovicff, line 12: This feature should be enabled only when we are testing. So probably add [dev-dependencies] and list copy the halo2 dep with this feature enabled

duguorong009: I tried to just copy the dependencies. 
But, it does not work since the feature should be enabled.

So, I just removed the `dev-graph` feature since it is for development.
https://github.com/eigen-trust/protocol/pull/169/commits/345abec5864f526141995961b9c92d770846d4c5

lazovicff: Ok, it can be removed if its not used in tests anymore. If you ever need it again in the future, we should have two version:
```toml
[dependencies]
halo2 = { package = "halo2_proofs", git = "https://github.com/privacy-scaling-explorations/halo2", tag = "v2023_02_02" }
// ...
[dev-dependencies]
halo2 = { package = "halo2_proofs", git = "https://github.com/privacy-scaling-explorations/halo2", tag = "v2023_02_02", features = ["dev-graph"] }
```

But for now its ok to exclude it

circuit/src/gadgets/mod.rs
```diff
@@ -8,3 +8,13 @@ pub mod lt_eq;
...
+
+/* NOTE:
+   Following chipsets are not used atm because of column limits.
+   Currently, the maximum columns for EigenTrust circuit is 8 advice + 5 fixed.
+   We plan to use them when lookup table can be used in circuit.
+*/
+// /// Check if a number is less than or equal to a value (using lookup table)
+// pub mod lt_eq_lookup;
+// /// Range check gadget (using lookup table)
+// pub mod range;
```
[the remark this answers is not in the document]
duguorong009: Okay, will update the `range` & `lt_eq_lookup` so that they use kind of mock config.

One question, do you think it is better to leave the `NOTE` comments or remove it after update? @lazovicff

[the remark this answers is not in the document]
lazovicff: I think the NOTE is very useful, so we should leave it there.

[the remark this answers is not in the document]
lazovicff: @duguorong009 Pushed a possible solution.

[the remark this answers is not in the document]
duguorong009: what a trick! :100: @lazovicff 
I think that works best.

circuit/src/gadgets/range.rs
```diff
@@ -0,0 +1,485 @@
...
+			//    z_i = 2^{K}⋅z_{i + 1} + a_i
+			// => a_i = z_i - 2^{K}⋅z_{i + 1}
+			let running_sum_lookup = {
+				let running_sum_word = {
+					let z_next = meta.query_advice(running_sum, Rotation::next());
+					z_cur.clone() - z_next * F::from(1 << K)
+				};
+
+				running_sum_selector.clone() * running_sum_word
+			};
```
lazovicff, line 141: For code readability, I think its very useful to avoid nesting

circuit/src/gadgets/range.rs
```diff
@@ -0,0 +1,485 @@
...
+							.collect::<Vec<_>>()
+					});
+
+					bits.map(|bits| {
+						bits.chunks_exact(K)
+							.map(|word| F::from(lebs2ip::<K>(&(word.try_into().unwrap()))))
+							.collect::<Vec<_>>()
+					})
+					.transpose_vec(N)
+				};
```
lazovicff, line 177: Also, can we avoid nested blocks

circuit/src/gadgets/range.rs
```diff
@@ -0,0 +1,485 @@
...
+
+					// z_next = (z_cur - m_cur) / 2^K
+					z = {
+						let z_val =
+							z.value().zip(word).map(|(z, word)| (*z - word) * inv_two_pow_k);
+
+						// Assign z_next
+						ctx.next();
+						ctx.assign_advice(mock_common.common.advice[0], z_val)?
+					};
```
lazovicff, line 201: Nested block

circuit/src/gadgets/range.rs
```diff
@@ -0,0 +1,485 @@
...
+			let words = bits
+				.map(|bits| {
+					bits.chunks_exact(K)
+						.map(|word| F::from(lebs2ip::<K>(&(word.try_into().unwrap()))))
+						.collect::<Vec<_>>()
+				})
+				.transpose_vec(N);
+
+			words[N - 1]
+		};
```
lazovicff, line 276: Nested block

circuit/src/utils.rs
```diff
@@ -291,3 +292,46 @@ where
...
+pub fn big_to_fe<F: FieldExt>(e: BigUint) -> F {
+	let modulus = modulus::<F>();
+	let e = e % modulus;
+	F::from_str_vartime(&e.to_str_radix(10)[..]).unwrap()
+}
+
+/// Conver the [`Field`] element to [`BigUint`] value
+pub fn fe_to_big<F: FieldExt>(fe: F) -> BigUint {
+	BigUint::from_bytes_le(fe.to_repr().as_ref())
+}
```
lazovicff, line 326: These are the copies from `integer/rns.rs`. I think this is the better place for them, so delete the ones from rns

circuit/src/utils.rs
```diff
@@ -291,3 +292,46 @@ where
 
 	Ok(res)
 }
+
+/// The u64 integer represented by an L-bit little-endian bitstring.
+///
+/// # Panics
+///
+/// Panics if the bitstring is longer than 64 bits.
+pub fn lebs2ip<const L: usize>(bits: &[bool; L]) -> u64 {
```
lazovicff, line 301: Can we call this `bits_to_u64` to stay consistent

circuit/src/gadgets/range.rs
duguorong009: Yes, I agree with you.
I updated all of nested blocks.
https://github.com/eigen-trust/protocol/pull/169/commits/ab695818a4f077629984c7e35e847639235712ca

circuit/src/gadgets/range.rs
duguorong009: Done https://github.com/eigen-trust/protocol/pull/169/commits/ab695818a4f077629984c7e35e847639235712ca

circuit/src/gadgets/range.rs
duguorong009: Done https://github.com/eigen-trust/protocol/pull/169/commits/ab695818a4f077629984c7e35e847639235712ca

circuit/src/gadgets/range.rs
duguorong009: Done https://github.com/eigen-trust/protocol/pull/169/commits/ab695818a4f077629984c7e35e847639235712ca

circuit/src/utils.rs
duguorong009: Yes, I done it here. https://github.com/eigen-trust/protocol/pull/169/commits/ab695818a4f077629984c7e35e847639235712ca

circuit/src/utils.rs
duguorong009: I add slight modification. 
I renamed `lebs2ip` to `le_bits_to_u64`.
I think `le` should be added for clearer name.
https://github.com/eigen-trust/protocol/pull/169/commits/ab695818a4f077629984c7e35e847639235712ca

circuit/src/utils.rs
```diff
@@ -291,3 +292,46 @@ where
...
+
+/// Get the little-endian bits array of [`Field`] element
+pub fn fe_to_le_bits<F: FieldExt>(e: F) -> Vec<bool> {
+	let le_bytes = fe_to_big(e).to_bytes_le();
+
+	let short_le_bytes = to_short(&le_bytes);
+
+	let le_bits: [bool; 256] = to_bits(short_le_bytes);
+
+	le_bits.to_vec()
```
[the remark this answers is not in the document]
duguorong009: Um, I tried your solution. And found that it does not work as expected.
The reason is that `to_bits` function is converting `le_bytes` to the `le_bits` in essence.
So, we should pass `le_bytes` data to the `to_bits` util function.

Simply, I just leave original function. Pls re-check it.

circuit/src/gadgets/range.rs
```diff
@@ -131,14 +131,11 @@ impl<F: FieldExt, const K: usize, const N: usize> MockChip<F> for LookupRangeChe
...
-				let running_sum_word = {
-					let z_next = meta.query_advice(running_sum, Rotation::next());
-					z_cur.clone() - z_next * F::from(1 << K)
-				};
-
-				running_sum_selector.clone() * running_sum_word
+			let running_sum_word = {
+				let z_next = meta.query_advice(running_sum, Rotation::next());
+				z_cur.clone() - z_next * F::from(1 << K)
 			};
```
lazovicff: This is still a nested block

duguorong009: Done https://github.com/eigen-trust/protocol/pull/169/commits/4cf0c213db9f7092b70767861e6e29ca630aad05

circuit/src/utils.rs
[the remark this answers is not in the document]
lazovicff: I make a mistake. The `reverse` should've been done on the array above - `short_bytes`.
We can leave it as is for now, just removed spaces in https://github.com/eigen-trust/protocol/pull/169/commits/2a2e5ea892a4ab8fc26ad4b605619a77d364c527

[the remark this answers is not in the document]
lazovicff: We usually cluster the lines based on context, and since these lines have a common purpose, its more readable to put them together

circuit/src/gadgets/range.rs
```diff
@@ -0,0 +1,471 @@
...
+					ctx.enable(selector.clone())?;
+
+					// z_next = (z_cur - m_cur) / 2^K
+					let z_next = z.value().zip(word).map(|(z, word)| (*z - word) * inv_two_pow_k);
+
+					// Assign z_next
+					ctx.next();
+					z = ctx.assign_advice(mock_common.common.advice[0], z_next)?;
+
+					zs.push(z.clone());
```
lazovicff, line 190: Do we need the whole array, just to get the last value from it?

circuit/src/gadgets/range.rs
```diff
@@ -0,0 +1,471 @@
...
+
+				// Assign shifted value
+				let shifted_word = self.x.value().cloned() * Value::known(F::from(1 << (K - S)));
+				ctx.assign_advice(mock_common.common.advice[0], shifted_word)?;
+				ctx.enable(selector.clone())?;
+
+				ctx.next();
+
+				// Assign 2^{-S} from a fixed column.
+				let inv_two_pow_s = F::from(1 << S).invert().unwrap();
```
lazovicff, line 92: The constant `1 << K` is defined as a contract expression: https://github.com/eigen-trust/protocol/pull/169/files#diff-11513d52971f0935b0b5cf0c36e0b15b6d9ef1595f05ed0915140a3cd5175dfbR58,  Why didnt you do the same with `inv_two_pow_s`

circuit/src/gadgets/lt_eq_lookup.rs
```diff
@@ -0,0 +1,381 @@
...
+	}
+}
+
+/// A chip for checking if number is in range
+struct LessEqualChipset<F: FieldExt> {
+	x: AssignedCell<F, F>,
+	y: AssignedCell<F, F>,
+	/// Bits of x and y and their difference
+	x_bits: [F; NUM_BITS],
+	y_bits: [F; NUM_BITS],
```
lazovicff, line 102: If they are not used, they should be deleted

circuit/src/gadgets/range.rs
```diff
@@ -0,0 +1,471 @@
...
+	pub fn new(x: AssignedCell<F, F>) -> Self {
+		Self { x }
+	}
+}
+
+impl<F: FieldExt, const K: usize, const N: usize> MockChip<F> for LookupRangeCheckChip<F, K, N> {
+	type Output = AssignedCell<F, F>;
+
+	fn configure(
+		mock_common: &MockCommonConfig, meta: &mut halo2::plonk::ConstraintSystem<F>,
```
lazovicff, line 120: We should import `ConstraintSystem` at the top of the file, instead of specifying the whole path:  `halo2::plonk::ConstraintSystem`
Do the same thing in every similar situation

duguorong009: Done https://github.com/eigen-trust/protocol/pull/169/commits/f5cb0f482c9f45b0327720226205386fe8a03613

circuit/src/gadgets/lt_eq_lookup.rs
duguorong009: Done https://github.com/eigen-trust/protocol/pull/169/commits/f5cb0f482c9f45b0327720226205386fe8a03613

circuit/src/gadgets/range.rs
duguorong009: Yes, I think we just need the last cell for zero check.
I removed the array logic from code. https://github.com/eigen-trust/protocol/pull/169/commits/f5cb0f482c9f45b0327720226205386fe8a03613

circuit/src/gadgets/range.rs
```diff
@@ -0,0 +1,471 @@
...
+			.transpose_vec(N);
+
+		// Take away the last word of `element`(x) & check if it is 4 bit
+		let last_word = words[N - 1];
+
+		let last_word_cell = layouter.assign_region(
+			|| "last word of x",
+			|region| {
+				let mut ctx = RegionCtx::new(region, 0);
+				ctx.assign_advice(mock_common.common.advice[0], last_word)
```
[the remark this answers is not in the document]
duguorong009: Yes, I thought about this.

I updated code so that the `LookupRangeCheckChip` returns both `x` and its "last word".
Here, the "last word" is a pre-ultimate element of `running_sum` check array.
Example: 
```
	// x:   1111111100000001
        // a_0: 11111111
        // a_1: 00000001
        //
        // z_0 = x
        // z_1 = (z_0 - a_0) / (2 ^ K)
        // z_2 = (z_1 - a_1) / (2 ^ K)
        //
        // [1111111100000001, 00000001, 0]
```
(In the example, the bits are little-endian ones. K = 8)
In this example, we can see that the "last word"(`00000001`) is the second-last element of `running_sum` check array.

circuit/src/gadgets/range.rs
```diff
@@ -0,0 +1,476 @@
...
+		)?;
+
+		let last_word_cell = layouter.assign_region(
+			|| "copy last word of x",
+			|region| {
+				let mut ctx = RegionCtx::new(region, 0);
+				ctx.copy_assign(mock_common.common.advice[0], last_word.clone())
+			},
+		)?;
+		let short_word_chip = LookupShortWordCheckChip::<F, K, S>::new(last_word_cell);
```
lazovicff, line 276: We can just pass `last_word` into the `LookupShortWordCheckChip` constructor.
The value will be copy assigned inside it

circuit/src/gadgets/range.rs
```diff
@@ -0,0 +1,476 @@
...
+
+impl<F: FieldExt, const K: usize, const N: usize, const S: usize> MockChipset<F>
+	for RangeChipset<F, K, N, S>
+{
+	type Config = RangeChipsetConfig;
+	type Output = AssignedCell<F, F>;
+
+	fn synthesize(
+		self, mock_common: &MockCommonConfig, config: &RangeChipsetConfig,
+		mut layouter: impl halo2::circuit::Layouter<F>,
```
lazovicff, line 259: Remove `halo2::circuit::`

circuit/src/gadgets/range.rs
lazovicff: This same constant is used in every row, so there is no need to store it inside the fixed column.
We can just hard code it into the constraint check.

circuit/src/gadgets/range.rs
duguorong009: Done https://github.com/eigen-trust/protocol/pull/169/commits/2b7cbfb7f9b4e8032e5c349a1e5f86dd98194146

circuit/src/gadgets/range.rs
duguorong009: Done https://github.com/eigen-trust/protocol/pull/169/commits/2b7cbfb7f9b4e8032e5c349a1e5f86dd98194146

circuit/src/gadgets/range.rs
```diff
@@ -0,0 +1,469 @@
...
+			vec![(bitshift_selector * shifted_word, mock_common.table)]
+		});
+
+		// For short lookups, check that the word has been shifted by the correct number
+		// of bits. https://p.z.cash/halo2-0.1:decompose-short-lookup
+		meta.create_gate("Short word S bits limit", |meta| {
+			let bitshift_selector = meta.query_selector(bitshift_selector);
+			let word = meta.query_advice(word_column, Rotation::prev());
+			let shifted_word = meta.query_advice(word_column, Rotation::cur());
+			let inv_two_pow_s = meta.query_advice(word_column, Rotation::next());
```
lazovicff, line 54: Why use 1 column and multiple rows, when there are available columns left unused.
The general idea when designing constraints is to use up all the columns in one row until there are none left, then move to the next row using rotations

lazovicff: For `LookupRangeCheckChip` this is hard to do, since we are calculating running sum for array of words (and array can have variable length). We would have to split the array into chunks, when the length of the chunk is the same as the amount of columns.
This will be hard to do, so we can leave it for the future.
But `ShortWordCheckChip` is very easy since we just need 2 or 3 columns

duguorong009: Yes, the `ShortWordCheckChip` needs just 2 columns if we hard-code the `S` value like `K`.

duguorong009: Done https://github.com/eigen-trust/protocol/pull/169/commits/a5db03110ec94caff1837ee9d9dda57ccfbac444

circuit/src/gadgets/range.rs
duguorong009: Done https://github.com/eigen-trust/protocol/pull/169/commits/a5db03110ec94caff1837ee9d9dda57ccfbac444

Please re-check the update.

circuit/src/gadgets/range.rs
```diff
@@ -0,0 +1,460 @@
...
+				z_0 1111111100000001 (little-endian)
+
+				z_1 00000001  		11111111 = 1111111100000001 - 00000001 * (000000001)
+									a_0      = z_0 				- z_1 	   * (1 << 8)
+				z_2 00000000   		00000001 = 00000001 - 0   * (000000001)
+									a_1 	 = z_1 		- z_2 * (1 << 8)
+
+				In the case of a running sum decomposition, we recover the word from
+				the difference of the running sums:
+					z_i = 2^{K}⋅z_{i + 1} + a_i
```
lazovicff, line 131: Can you elaborate on how you got this equation.
Because [here](https://github.com/eigen-trust/protocol/pull/169/files#diff-11513d52971f0935b0b5cf0c36e0b15b6d9ef1595f05ed0915140a3cd5175dfbR175), the equation is:
`z_i = (z_{i - 1} - a_{i - 1}) / 2^K`

lazovicff (approved):

PR #170, opened 2023-03-07
Test fixes

lazovicff:


4rmanc (dismissed):

duguorong009 (dismissed):
Thanks!  :rocket:

duguorong009 (approved):

lazovicff:
@duguorong009 Can you re-review this? I think its ready now

duguorong009 (commented):
First of all, thanks for fixes! :+1: 
I checked the branch in my local machine & it works! :raised_hands: 

One question from curiosity.
You moved the `foundry(anvil)` install script from local to github CI.
Do you want the tests(related to `anvil`) to run in only CI, not local machines?

If so, what should I do for running those tests on local machine? just install `anvil`?

duguorong009 (approved):

lazovicff:
The tests are meant to run both locally and on CI.
The scripts inside ./scripts folder are only meant for Ubuntu linux which is used inside CI.
Locally, we should manually install all dependencies.
Here are the instructions for foundry:
https://github.com/foundry-rs/foundry#installation

duguorong009:
> The tests are meant to run both locally and on CI. The scripts inside ./scripts folder are only meant for Ubuntu linux which is used inside CI. Locally, we should manually install all dependencies. Here are the instructions for foundry: https://github.com/foundry-rs/foundry#installation

okay, thanks for the info! :+1:

PR #185, opened 2023-04-04
Test cases for Loader, Transcript and Aggregator

4rmanc:


circuit/src/verifier/transcript/mod.rs
```diff
@@ -354,46 +366,497 @@ mod test {
...
+				IntegerMulChip::<Base, Scalar, NUM_LIMBS, NUM_BITS, P>::configure(&common, meta);
+			let int_div =
+				IntegerDivChip::<Base, Scalar, NUM_LIMBS, NUM_BITS, P>::configure(&common, meta);
+
+			let ladder = EccUnreducedLadderConfig::new(int_add, int_sub, int_mul, int_div);
+			let add = EccAddConfig::new(int_red, int_sub, int_mul, int_div);
+			let double = EccDoubleConfig::new(int_red, int_add, int_sub, int_mul, int_div);
+			let table_select = EccTableSelectConfig::new(main.clone());
+			let ecc_mul_scalar = EccMulConfig::new(ladder, add, double, table_select, bits2num);
+			TestConfig { common, main, poseidon_sponge, ecc_mul_scalar }
```
lazovicff, line 794: I see that this duplicate in every test circuit.
Perhaps you could add a constructor for TestCircuit that does this

circuit/src/verifier/loader/mod.rs
```diff
@@ -629,7 +629,6 @@ where
 			.iter()
 			.cloned()
 			.map(|(scalar, base)| {
-				// TODO: Test stucks here, fix it.
 				let config = pairs[0].0.loader.clone();
 				let auxes = pairs[0].1.loader.auxes.clone();
```
lazovicff, line 633: We should take the loader from the `base` or `scalar`

lazovicff:
Instead of locking and droping the loader every time you use it, you could create a method for loader called `layouter` or similar to handle that for you:
```rust
fn layouter(func: FnOnce(layouter)) {
  let layouter = self.layouter.lock();
  func(layouter);
  drop(layouter);
}
```

lazovicff (approved):

Issue #190, opened 2023-04-06
Client README


lazovicff:
Add a readme that explains what are the dependencies for running the client and how to run it.
This document should serve as a starting point for developers trying to use it.
Should be based on this hackmd:
https://hackmd.io/c6nwvG3_TMa3TLLdvxIGNQ

PR #191, opened 2023-04-06
docs: client readme

brech1:
# Description

Completes #190 

Creates a README.md file for the protocol client. It's main intent being helping contributors to spin up the project and get an idea of how it works.

### DIFF
--- a/client/README.md
+++ b/client/README.md
@@ -1,0 +1,86 @@
+# EigenTrust Protocol
+
+This crate is an implementation of the EigenTrust Protocol. It's intended use is either as a library for other implementations or as a standalone implementation.
+
+## Requirements
+
+- Rust: To install, follow the instructions found [here](https://www.rust-lang.org/tools/install).
+- Solidity Compiler: To install, follow the instructions found [here](https://docs.soliditylang.org/en/v0.8.9/installing-solidity.html).
+- Anvil: CLI tool for running a local EVM blockchain. Follow these steps to install Anvil from source:
+
+```bash
+git clone https://github.com/foundry-rs/foundry
+cd foundry
+cargo install --path ./anvil --bins --locked --force
+```
+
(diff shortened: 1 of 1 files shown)


lazovicff:
Looks good.
I think it would be more readable if we tried to avoid nested blocks and too many if branching.
That's your take on this?

brech1:
> Looks good. I think it would be more readable if we tried to avoid nested blocks and too many if branching. Instead do something like:
> 
> ```rust
> if var.is_none() {
>   return;
> }
> let var = var.unwrap();
> ```
> 
> or similar cases.
> 
> That's your take on this?

I agree! there's quite some nesting going on. I'll submit a fix for that.

client/src/main.rs
```diff
@@ -93,84 +110,102 @@ async fn main() {
...
+
+			config.ops[pos] = score_parsed.unwrap();
 		},
-		Mode::Show => {
-			println!("Client config:");
-			println!("{:#?}", config);
+		Config::SecretKey => {
+			let sk_vec: Vec<String> = data.split(",").map(|x| x.to_string()).collect();
+			if sk_vec.len() != 2 {
+				return Err("Invalid secret key passed, expected 2 bs58 values separated by commas, e.g.:\n\"2L9bbXNEayuRMMbrWFynPtgkrXH1iBdfryRH9Soa8M67,9rBeBVtbN2MkHDTpeAouqkMWNFJC6Bxb6bXH9jUueWaF\"".to_string());
```
lazovicff, line 190: Can we break this into several lines?
Its best to add `format_strings = true` inside `rustfmt.toml` file..

brech1: Solved, added `format_strings = true` to `rustfmt.toml`

lazovicff (approved):

Issue #186, opened 2023-04-06
Merge client & server


lazovicff:
The goal is to simplify the protocol by merging client and the server crates:
1) Copy the manager logic into client - separate folder
2) Copy `MANAGER_STORE` lazy static into the client
3) Use the manager for computing proof instead of calling the server API

Cleanup afterwards:
1) Remove server crate
2) Remove all references to the server in the workspace
3) Remove unused dependencies like reqwest.rs

PR #193, opened 2023-04-11
Merge client & server WIP

brech1:
# Description

Solves #186

The aim of this PR is to simplify the protocol and merge the client and server crates.

The proof generation will be embedded in the client CLI.


lazovicff:
Is it possible for you to switch to main repo, instead of working from fork.
For now, you can create a branch on main repo and merge this code into it.

client/src/manager.rs
```diff
@@ -208,32 +225,19 @@ impl Manager {
 		// --- END ---
```
lazovicff, line 225: Maybe we should move this into the test `should_calculate_proof`, instead of doing sanity checks

brech1: Sure! Makes sense, the snippet only runs on unoptimized builds

brech1:
> Is it possible for you to switch to main repo, instead of working from fork? For now, you can create a branch on main repo and merge this code into it.

Yes, I just pushed some updates and we can merge this to the new branch on the main repo.

PR #195, opened 2023-04-13
Merge client & server

brech1:
# Description

This PR simplifies the project structure and establishes a new way of interacting with the protocol. 

- Solves #186

## Changes

- The `server` crate has been removed. The project no longer has a HTTP server implementation.
- The attestation proof generation functionality has now been moved to the `client` directory.
- There is a new client CLI command: `generate-proof`. This will fetch every attestation from the contract, generate a proof and store it into a file.
- The `verify` command has now been updated to load the proof from the file instead of fetching it from the server.
- The README has been updated to describe this new functionality.
- It also include some minor fixes.


client/README.md
```diff
@@ -75,17 +75,8 @@ The client configuration is stored in `data/client-config.json`, which specifies
...
-
-### Server
-
-The server was built using [hyper.rs](http://hyper.rs/). You can find the configuration file at `data/protocol-config.json`, where you can specify the following:
-
-- `epoch_interval`: Interval at which proofs are calculated.
-- `endpoint`: Socket that listens for connections to the server.
 - `ethereum_node_url`: URL of the Ethereum node we are connecting to. This defaults to `127.0.0.1:8545` to run with a local `anvil` EVM blockchain.
-- `as_contract_address`: Address of the AttestationStation smart contract from which events are being fetched.
+- `server_url`: The URL for the running server.
```
lazovicff, line 82: This is not used

client/src/main.rs
```diff
@@ -99,15 +104,35 @@ async fn main() {
...
+				},
+			}
+
 			let client = EigenTrustClient::new(config, user_secrets_raw);
-			client.verify(proof_raw).await.unwrap();
-			println!("Successful verification!");
+
+			client.verify(ProofRaw::from(proof)).await.unwrap();
+			println!("Proof verified");
 		},
```
lazovicff, line 136: Maybe we should split this into two commands:
- Generate proof and save to file
- Read proof from the file and verify it

We should go in this direction because we will have proofs for multiple different sets that we want to aggregate in different ways

lazovicff: Also, we could get rid of some nesting

client/src/manager.rs
```diff
@@ -70,8 +75,8 @@ pub const PUBLIC_KEYS: [&str; NUM_NEIGHBOURS] = [
 
 /// The peer struct.
 pub struct Manager {
-	pub(crate) cached_proofs: HashMap<Epoch, Proof>,
-	pub(crate) attestations: HashMap<Scalar, Attestation>,
+	attestations: HashMap<Scalar, Attestation>,
+	cached_proofs: Vec<Proof>,
 	params: ParamsKZG<Bn256>,
 	proving_key: ProvingKey<G1Affine>,
 	verifier_code: Vec<u8>,
```
lazovicff, line 82: I think this is no longer needed here

lazovicff:
Whats the reasoning behind `attestation.rs` inside `manager` folder. The Attestation is used both in Manager and the EigenTrustClient.

Btw, we could rename `EigenTrustClient` to `Client`.

client/src/utils.rs
```diff
@@ -150,13 +158,62 @@ pub fn compile_yul_contracts() {
...
+
+/// Get the attestations from the contract
+pub async fn get_attestations(config: &ClientConfig) -> Result<Vec<Attestation>, EigenError> {
+	let client = setup_client(&config.mnemonic, &config.ethereum_node_url);
+	let contract = AttestationStation::new(config.as_address.parse::<Address>().unwrap(), client);
+	let events = contract.event::<AttestationCreatedFilter>().from_block(0);
+	let mut stream = events.stream().await.unwrap();
+	let mut attestations = Vec::new();
+
+	while let Ok(Some(Ok(att_created))) = timeout(Duration::from_secs(10), stream.next()).await {
```
lazovicff, line 202: Timeouts are not ideal. Is there a way to fetch all historic events instead of streaming them?
I did some quick research, perhaps there is a solution here:
https://github.com/gakonst/ethers-rs/issues/988 or
https://github.com/gakonst/ethers-rs/blob/master/examples/events/examples/filtering.rs

brech1: This has been very helpful! Thanks, just submitted a fix for it using `logs` instead of a `stream`

client/src/manager.rs
brech1: True, I removed the `verifier_code`

client/README.md
brech1: Thanks. I just removed the `server_url` from here and the `client-config.json`. I'm still waiting to have the CLI functionality defined to update that part of the readme.

brech1:
> Whats the reasoning behind `attestation.rs` inside `manager` folder. The Attestation is used both in Manager and the EigenTrustClient?
> 
> Btw, we could rename `EigenTrustClient` to `Client`.

It was inherited from the server structure, but it makes sense for `attestation` to be it's own module. I'll submit a fix for this, including the renaming of `EigenTrustClient`.

client/src/main.rs
brech1: Sure! I kept this in the same action mainly for quick prototyping.

I will implement this to save the latest proof in a file, and that the verify command reads from it and verifies it. However, I think we should discuss wether in the future we save multiple proofs in one file or separate files, and how do we specify which proof to verify.

lazovicff: This is yet to be discussed, but roughly every set will have its unique identifier which is going to be used as a file name as well along with a prefix that specifies if its a single set proof or multiple sets aggregated.

For now we can treat the whole network as a single proof and name the file `et_proof.json` or similar.

lazovicff (approved):

lazovicff (approved):

PR #196, opened 2023-04-16
Algorithm updates

lazovicff:


lazovicff:
While writing the docs, I noticed that we did our algorithm wrong.
This should be our reference: https://eigen-books.gitbook.io/eigen-trust/~/revisions/EzbPDwFJ1NWlnID22Uvc/4_algorithm

@duguorong009 Can you fix the failing test cases?

duguorong009:
> While writing the docs, I noticed that we did our algorithm wrong. This should be our reference: https://eigen-books.gitbook.io/eigen-trust/~/revisions/EzbPDwFJ1NWlnID22Uvc/4_algorithm
> 
> @duguorong009 Can you fix the failing test cases?

@lazovicff 
I checked the doc and algorithm updates, and I understand the intention.

While checking the failing test cases, I found that the `normalize` computation does not match the original intention.
For example, `[0, 10, 20, 10]` becomes `[0, 0.25, 0.5. 0.5]` when we follow the doc.
But, real computation is done in finite field. So, `[0, 10, 20, 10]` becomes `[0, 10 * 40^-1, 20 * 40 ^-1 , 10 * 40^-1]`.
Clearly, this result is not equal to decimal, and score computation is not equal.
Hence, after every iteration of computation, the sum would be different to previous sum.
As result, I think it is not correct to add the constraint `sum_initial == sum_final` [here](https://github.com/eigen-trust/protocol/blob/native-fix/circuit/src/native.rs#L136)

lazovicff:
For now we can comment out `sum_initial == sum_final` assert, and review & merge this

duguorong009 (approved):

Issue #192, opened 2023-04-07
Dynamic sets circuit updates


lazovicff:
Create a copy of EigenTrust circuit and add support for dynamic set filtering: #141 

Also we should create separate folder called `dynamic_sets` that contains the native and Halo2 version of dynamic sets:
- dynamic_sets/
  - mod.rs
  - native.rs

The requirements is OrChipset.

PR #197, opened 2023-04-17
Dynamic sets circuit updates

duguorong009:
## Description
ZK circuit for `dynamic-sets` algorithm is implemented in this PR.
 - Solves #192 
 - Solves #198 (partially, only sum check part)

## Changes
 - Create `dynamic-sets` dir for both native and circuit code of `Dynamic Sets`
 - Implement the `EigenTrustSet` circuit 
 - Add the unit tests for circuit
 - Some minor updates to native version
 - Fix the EigenTrust algo implementation in native version
 - Update the `InverseChipset` so that it can handle even un-invertible value
 - Add `OrChipset`(logical `or`) implementation

### DIFF
--- a/circuit/src/dynamic_sets/mod.rs
+++ b/circuit/src/dynamic_sets/mod.rs
@@ -653,10 +653,10 @@ impl<const NUM_NEIGHBOURS: usize, const NUM_ITER: usize, const INITIAL_SCORE: u1
 		// Compute the EigenTrust scores
 		let mut s = vec![init_score.clone(); NUM_NEIGHBOURS];
 		for _ in 0..NUM_ITER {
-			let mut distributions = Vec::new();
+			let mut sop = Vec::new();
 			for i in 0..NUM_NEIGHBOURS {
 				let op_i = ops[i].clone();
-				let mut local_distr = Vec::new();
+				let mut sop_i = Vec::new();
 				for j in 0..NUM_NEIGHBOURS {
 					let mul_chip = MulChipset::new(op_i[j].clone(), s[i].clone());
 					let res = mul_chip.synthesize(
@@ -664,15 +664,15 @@ impl<const NUM_NEIGHBOURS: usize, const NUM_ITER: usize, const INITIAL_SCORE: u1
 						&config.main,
 						layouter.namespace(|| "op_mul"),
 					)?;
-					local_distr.push(res);
+					sop_i.push(res);
 				}
-				distributions.push(local_distr);
+				sop.push(sop_i);
 			}
 
 			let mut new_s = vec![zero.clone(); NUM_NEIGHBOURS];
 			for i in 0..NUM_NEIGHBOURS {
 				for j in 0..NUM_NEIGHBOURS {
-					let add_chip = AddChipset::new(new_s[i].clone(), distributions[j][i].clone());
+					let add_chip = AddChipset::new(new_s[i].clone(), sop[j][i].clone());
 					new_s[i] = add_chip.synthesize(
 						&config.common,
 						&config.main,
--- a/circuit/src/dynamic_sets/mod.rs
+++ b/circuit/src/dynamic_sets/mod.rs
@@ -326,285 +326,285 @@ impl<const NUM_NEIGHBOURS: usize, const NUM_ITER: usize, const INITIAL_SCORE: u1
 			},
 		)?;
 
-		// // signature verification
-		// for i in 0..NUM_NEIGHBOURS {
-		// 	let mut pk_sponge = SpongeHasher::new();
-		// 	pk_sponge.update(&op_pk_x[i]);
-		// 	pk_sponge.update(&op_pk_y[i]);
-		// 	let pks_hash = pk_sponge.synthesize(
-		// 		&config.common,
-		// 		&config.sponge,
-		// 		layouter.namespace(|| "pks_sponge"),
-		// 	)?;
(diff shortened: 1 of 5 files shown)


circuit/src/dynamic_sets/mod.rs
```diff
@@ -0,0 +1,490 @@
...
+
+		// filter peers' ops
+		let ops = {
+			let mut filtered_ops = Vec::new();
+			for i in 0..NUM_NEIGHBOURS {
+				let mut ops_i = ops[i].clone();
+
+				// Update the opinion array - pairs of (key, score)
+				//
+				// TODO: How to nullify the op_score according to pair of (set_pk, op_pk)
```
lazovicff, line 347: We select `0` or `org_score` based on some condition using `SelectChipset`.

For example, if the condition is unequal keys:
```rust
let is_equal_pk = IsEqaulChipset::new(set_pk, curr_pk);
let maybe_nullified_score = SelectChipset(is_equal_pk, org_score, zero);
// Mutate the opinion array at index `i` with `maybe_nullified_score`
```

circuit/src/dynamic_sets/mod.rs
```diff
@@ -0,0 +1,490 @@
...
+				}
+
+				// Add to "filtered_ops"
+				filtered_ops.push(ops_i);
+			}
+
+			filtered_ops
+		};
+
+		// Maybe, need to do "normalize" here?
```
lazovicff, line 407: We have `InvertChipset`, so we do the same as in native version:
```rust
let inverted_sum = InvertChipset::new(sum);
let normalised_score = MulChipset::new(score, inverted_sum);
```

circuit/src/dynamic_sets/mod.rs
```diff
@@ -0,0 +1,878 @@
...
+					if is_diff_pk_j || is_pk_j_null || is_pk_i {
+						ops_i[j] = zero.clone();
+					}
+
+					// Condition for correcting the pk
+					// 1. set_pk_j != op_pk_j
+					if is_diff_pk_j {
+						op_pk_x[i][j] = set_pk_j_x;
+						op_pk_y[i][j] = set_pk_j_y;
+					}
```
lazovicff, line 466: We cant use if statements here, since this would mean that the circuit layout is non deterministic.
We should use SelectChipset when nullifying pks and scores.

circuit/src/dynamic_sets/mod.rs
```diff
@@ -0,0 +1,878 @@
...
+								&config.common,
+								&config.main,
+								layouter.namespace(|| "pk_j_y == default_pk_y"),
+							)
+							.is_ok();
+						let is_not_null = !(is_default_pk_x && is_default_pk_y);
+
+						if is_diff_pk && is_not_null {
+							ops_i[j] = one.clone();
+						}
```
lazovicff, line 515: Should use SelectChipset here as well

circuit/src/dynamic_sets/mod.rs
```diff
@@ -0,0 +1,961 @@
...
+			let mut passed_scaled = Vec::new();
+			for i in 0..NUM_NEIGHBOURS {
+				let mul_chip = MulChipset::new(passed_s[i].clone(), scale.clone());
+				let res = mul_chip.synthesize(
+					&config.common,
+					&config.main,
+					layouter.namespace(|| "op_mul"),
+				)?;
+				passed_scaled.push(res);
+			}
```
lazovicff, line 708: This is not relevant anymore since we are normalising values

circuit/src/dynamic_sets/mod.rs
```diff
@@ -0,0 +1,961 @@
...
+						new_s[i] = add_chip.synthesize(
+							&config.common,
+							&config.main,
+							layouter.namespace(|| "op_add"),
+						)?;
+					}
+				}
+
+				s = new_s;
+			}
```
lazovicff, line 697: We should simplify this to match the one from the docs

circuit/src/dynamic_sets/mod.rs
```diff
@@ -0,0 +1,912 @@
...
+				// Compute the sum of scores
+				let mut op_score_sum = zero.clone();
+				for j in 0..NUM_NEIGHBOURS {
+					let add_chip = AddChipset::new(op_score_sum.clone(), ops[i][j].clone());
+					op_score_sum = add_chip.synthesize(
+						&config.common,
+						&config.main,
+						layouter.namespace(|| "op_score_sum"),
+					)?;
+				}
```
lazovicff, line 615: The sum is already calculated once here: https://github.com/eigen-trust/protocol/pull/197/files#diff-d1a67c0b28a8f922a0aafb9671a55106124235c920ec65ab43c1ea9f883b67b1R494-R502

Perhaps we could reuse it

duguorong009: Hmm, I don't think we can reuse it.
The `op_score_sum` could change since there is logic of assigning `Fr::one()` to opinion. (https://github.com/eigen-trust/protocol/blob/dynamic-sets-circuit/circuit/src/dynamic_sets/mod.rs#L569-L590).
Hence, I think we should re-compute the `op_score_sum`.

lazovicff: Right, sorry I overlooked that.

lazovicff (approved):

PR #200, opened 2023-04-19
Run cargo update

lazovicff:


brech1:
I had to reinstall rust because I had a compile error with my previous version.

When running cargo update I get some changes:

```toml
name = "clap"
version = "4.2.4"

##

name = "clap_builder"
version = "4.2.4"

##

name = "linux-raw-sys"
version = "0.3.3"
```

Tests and build run ok.

lazovicff:
> I had to reinstall rust because I had a compile error with my previous version.
> 
> When running cargo update I get some changes:
> 
> ```toml
> name = "clap"
> version = "4.2.4"
> 
> ##
> 
> name = "clap_builder"
> version = "4.2.4"
> 
> ##
> 
> name = "linux-raw-sys"
> version = "0.3.3"
> ```
> 
> Tests and build run ok.

Thanks, updated Rust version.
I am not sure why your `cargo` doesn't just pick up the version from `rust-toolchain.toml`.

brech1 (approved):

lazovicff:
@brech1 Using stable rust, but wasn't able to fully switch to new version of halo2. The reason is that `snark-verifier` (our dependency) uses older version of halo2 (https://github.com/privacy-scaling-explorations/snark-verifier/blob/main/snark-verifier/Cargo.toml#L20), which makes the types incompatible

brech1 (approved):

PR #204, opened 2023-04-27
Snark witness

4rmanc:


circuit/src/verifier/aggregator.rs
```diff
@@ -104,11 +97,51 @@ impl Snark {
...
+
 struct Aggregator {
 	// Succinct Verifying Key
 	svk: SVK,
 	// Snarks for the aggregation
-	snarks: Vec<Snark>,
+	snarks: Vec<SnarkWitness>,
 	// Instances
 	instances: Vec<Fr>,
 	// Accumulation Scheme Proof
```
lazovicff, line 147: While we are here, I think `as_proof` should be wrapped by `Value` as well

circuit/src/verifier/transcript/mod.rs
```diff
@@ -136,29 +136,23 @@ where
...
-		})?;
+		let scalar = self.reader.as_mut().and_then(|reader| {
+			let mut data = <C::Scalar as PrimeField>::Repr::default();
+			if reader.read_exact(data.as_mut()).is_err() {
+				return Value::unknown();
+			}
+			Option::<C::Scalar>::from(C::Scalar::from_repr(data))
+				.map(Value::known)
+				.unwrap_or_else(Value::unknown)
+		});
```
lazovicff, line 147: We should leave error handling here as we did before.
If we fail to read from the reader, we shouldn't return `Value::unknown()`, instead we should return error.

circuit/src/verifier/transcript/mod.rs
```diff
@@ -172,24 +166,21 @@ where
...
+			if reader.read_exact(compressed.as_mut()).is_err() {
+				return Value::unknown();
+			}
+			let coords = C::from_bytes(&compressed).unwrap().coordinates().unwrap();
+			x = Some(Integer::<_, _, NUM_LIMBS, NUM_BITS, P>::from_w(*coords.x()));
+			y = Some(Integer::<_, _, NUM_LIMBS, NUM_BITS, P>::from_w(*coords.y()));
+			Option::<C>::from(C::from_bytes(&compressed))
+				.map(Value::known)
+				.unwrap_or_else(Value::unknown)
+		});
```
lazovicff, line 183: Here as well.
We should handle errors properly

4rmanc:
@lazovicff in [here](https://github.com/eigen-trust/protocol/blob/snark-witness/circuit/src/verifier/aggregator.rs#L301-L311) when I use `ctx.assign_advice()` function it returns `None` value. If I put correct instances manually, `Value::known(Fr::one())` for example, it still returns `None`. I couldn't find the reason behind it. What can cause it? It works when I use `assign_from_constant()` function.

circuit/src/verifier/transcript/mod.rs
```diff
@@ -173,24 +174,28 @@ where
...
-				ErrorKind::Other,
-				"invalid point encoding in proof".to_string(),
-			)
-		})?;
-		let coordinates = point.coordinates().unwrap();
+		let _ = self.reader.as_mut().and_then(|reader| {
+			let mut compressed = C::Repr::default();
+			if reader.read_exact(compressed.as_mut()).is_err() {
+				return Value::unknown();
+			}
```
lazovicff, line 184: Handle this error

circuit/src/verifier/transcript/mod.rs
```diff
@@ -137,29 +137,30 @@ where
...
-			VerifierError::Transcript(
-				err.kind(),
-				"invalid field element encoding in proof".to_string(),
-			)
-		})?;
+		let scalar = self.reader.as_mut().and_then(|reader| {
+			let mut data = <C::Scalar as PrimeField>::Repr::default();
+			if reader.read_exact(data.as_mut()).is_err() {
+				return Value::unknown();
+			}
```
lazovicff, line 144: We should handle this error

circuit/src/verifier/transcript/mod.rs
```diff
@@ -173,24 +174,28 @@ where
...
+			x = Some(Integer::<_, _, NUM_LIMBS, NUM_BITS, P>::from_w(*coords.x()));
+			y = Some(Integer::<_, _, NUM_LIMBS, NUM_BITS, P>::from_w(*coords.y()));
+			let value = Option::<C>::from(C::from_bytes(&compressed))
+				.ok_or_else(|| {
+					VerifierError::Transcript(
+						ErrorKind::Other,
+						"invalid field element encoding in proof".to_string(),
+					)
+				})
+				.unwrap();
```
lazovicff, line 195: We should not `unwrap`, but instead propagate error down the stack

4rmanc: I pushed a new commit and I tried manual printing for handling it. Is it good way to handle it manually by printing error and returning unknown() value? Value type doesn't allow me to handle it in traditional way.

circuit/src/verifier/transcript/mod.rs
4rmanc: Because `Reader` type is wrapped in `Value` type, it doesn't allow me to return error. Instead I can manually print the error if it returns error inside and then return `Value::unknown()`. If I carry return inside the `Value` type it gives type errors when I try to assign them to a region.

4rmanc:
I approached from another perspective and found a way to work it out. I took reader out from `Value` type and did the rest of the calculations as we did before. At the end I sent it to the function by wrapping `Value::known()`

circuit/src/verifier/transcript/mod.rs
```diff
@@ -137,20 +137,27 @@ where
...
+		// Taking out reader from Value for a proper error handling
+		let mut reader_out: Option<&mut RD> = None;
+		let _ = self.reader.as_mut().and_then(|reader| {
+			reader_out = Some(reader);
+			Value::known(<C::Scalar as PrimeField>::Repr::default())
+		});
+
 		let mut data = <C::Scalar as PrimeField>::Repr::default();
-		self.reader.read_exact(data.as_mut()).map_err(|err| {
+		reader_out.unwrap().read_exact(data.as_mut()).map_err(|err| {
```
lazovicff, line 148: We cant have unwrap here. It will panic if `reader` is `Value::unknown`

lazovicff:
> I approached from another perspective and found a way to work it out. I took reader out from `Value` type and did the rest of the calculations as we did before. At the end I sent it to the function by wrapping `Value::known()`

I see. The `Value` type is making a lot of issues. I will look into this as well

lazovicff (approved):

Issue #198, opened 2023-04-18
Un-inverted scores


lazovicff:
We have to understand the differences between decimal and finite field computations, and add a constraint `sum_initial == sum_final`, to check against reputation leaking inside the EigenTrust algorithm.

We should also think about how to convert final s values into their un-inverted form, for them to be useful in real world.

More details about how the algorithm works: https://github.com/eigen-trust/docs/blob/master/4_algorithm.md

PR #205, opened 2023-04-27
Un-inverted scores

duguorong009:
## Description
Implement the scaling idea in `dynamic_sets` in order to convert large final scores to small range values(more readable)
  - Solves #198 

## Changes
  - Add the `scale_factor` idea in `native` version

## NOTE
Sum check part has been done in #192


circuit/src/dynamic_sets/native.rs
```diff
@@ -563,36 +687,174 @@ mod test {
...
+	}
+
+	#[test]
+	fn test_scaling_1() {
+		const NUM_NEIGHBOURS: usize = 2_usize.pow(10);
+		const NUM_ITERATIONS: usize = 20;
+		const INITIAL_SCORE: u128 = 1000000;
+
+		let real_neighbors = 2;
+		let is_random_score = true;
```
lazovicff, line 798: Why do we have a case where scores are not random?

lazovicff (approved):

PR #211, opened 2023-05-08
multi multiplication

saiakilesh:


lazovicff:
https://github.com/eigen-trust/protocol/issues/212 assumes that we are using the scalar from the wrong curve.
Just adding generic `<S: FieldExt>` is not enough.

It should be:
```rust
pub fn multi_mul_scalar(
  points: &[Self], scalars: &[Integer<C::Scalar, N, ...>], sliding_window_usize: usize,
) -> Vec<Self> {
```

But it should be done in separate file, more details in the issue itself

lazovicff:
Values like `aux_init`, `aux_fin`, `sliding_window_usize` and `num_of_windows` should be in RnsParams

saiakilesh:
@lazovicff I've addressed your comments. Should be ready for review.

lazovicff:
`circuit/src/ecc/native_general_ecc.rs` looks better. We should update `circuit/src/ecc/native.rs` to match it

circuit/src/rns/secp256k1.rs
```diff
@@ -85,36 +85,56 @@ impl RnsParams<Fp, Fr, 4, 68> for Secp256k1_4_68 {
...
-		let limb2 = Fr::from_u128(147045207087024714823);
-		let limb3 = Fr::from_u128(661255989794087);
+		let limb0 =
+			Fr::from_u128(0x00000000000000000000000000000000000000000000000a30efcc2747d66b3d);
+		let limb1 =
+			Fr::from_u128(0x00000000000000000000000000000000000000000000000871230a93e8f70fec);
+		let limb2 =
+			Fr::from_u128(0x0000000000000000000000000000000000000000000000032aa8e8ae3d48867b);
+		let limb3 =
+			Fr::from_u128(0x000000000000000000000000000000000000000000000000000651634b333363);
```
lazovicff, line 95: Why were these updated?

saiakilesh: I had a bug in `make_mul_aux` function. I edited it so the values changed.

saiakilesh:
@lazovicff What should the `sliding_window_usize` in `RnsParams`. There are multiple options: the sliding window could correspond to the bit size of `N` or `W`. Right now I have it corresponding to `W` which makes sense in the case of `generic.rs`. But does not make sense in the case of `native.rs` since scalar is from `N`. Hence I left `sliding_window_usize` as an input to the function. Let me know your thoughts. 

> Values like `aux_init`, `aux_fin`, `sliding_window_usize` and `num_of_windows` should be in RnsParams

lazovicff (approved):

PR #218, opened 2023-05-11
Un inverted scores 2

lazovicff:


circuit/src/dynamic_sets/native.rs
```diff
@@ -868,23 +880,29 @@ mod test {
...
 		const NUM_NEIGHBOURS: usize = 10;
 		const NUM_ITERATIONS: usize = 30;
 		const INITIAL_SCORE: u128 = 1000;
-		const SCALE: u128 = 100000000000000000000000000000000000000;
+		// Constants related to threshold check
+		const NUM_LIMBS: usize = 2;
+		const POWER_OF_TEN: usize = 50;
 
-		let ops = [
+		let ops_raw = [
```
duguorong009, line 905: Here, the `ops` is simulated one, I think.
I mean the sum of one peer's opinion is the same among peers. (1000)
Have you tried any test of random opinion scores where the sum is different between peers?

lazovicff: Yes, I tried, this just means we get much higher numbers for numerator and denominator as the result from `converge_rational`.
This would also mean that we have to use different params for `NUM_LIMBS` and `POWER_OF_TEN`. But to solve this, we will run the experiments with random values until we figure out what is the highest number of decimals we could get for numerator and denominator. Based on that we will scale our values and make some hardcoded params.
I will add this as TODO, since I need to focus on aggregators.

duguorong009 (approved):

duguorong009 (approved):

Issue #207, opened 2023-04-28
UnassignedValue Trait


lazovicff:
```rust
trait UnassignedValue {
  fn without_witnesses(&self) -> Self;
}
```

Example:
```rust
struct UnassignedPoint<F: FieldExt> {
  x: Value<F>,
  y: Value<F>
}

impl<F: FieldExt> UnassignedValue for UnassignedPoint<F> {
  fn without_witnesses(&self) -> Self {
    Self { x: Value::unknown(), y: Value::unknown() }
  }
}
```

- Implement this trait for:
  - PublicKey
  - Signature

- Update DynamicSets circuit to use these traits inside `without_witnesses` method.
- Update every test circuit defined inside unit test modules.

4rmanc:
@lazovicff What do you mean by saying unit test modules? You mentioned that I should implement this for all the test circuits in PR. Is there anything specific?

lazovicff:
https://github.com/eigen-trust/protocol/blob/master/circuit/src/ecc/mod.rs#L1046-L1048
https://github.com/eigen-trust/protocol/blob/master/circuit/src/ecc/mod.rs#L1121-L1123
https://github.com/eigen-trust/protocol/blob/master/circuit/src/ecc/mod.rs#L1193-L1195
https://github.com/eigen-trust/protocol/blob/master/circuit/src/ecc/mod.rs#L1270-L1272
https://github.com/eigen-trust/protocol/blob/master/circuit/src/eddsa/mod.rs#L242-L244
https://github.com/eigen-trust/protocol/blob/master/circuit/src/gadgets/bits2num.rs#L150-L152
https://github.com/eigen-trust/protocol/blob/master/circuit/src/gadgets/lt_eq.rs#L210-L212
https://github.com/eigen-trust/protocol/blob/master/circuit/src/gadgets/lt_eq_lookup.rs#L200-L202
https://github.com/eigen-trust/protocol/blob/master/circuit/src/gadgets/main.rs#L734-L736
https://github.com/eigen-trust/protocol/blob/master/circuit/src/gadgets/main.rs#L842-L844
https://github.com/eigen-trust/protocol/blob/master/circuit/src/gadgets/main.rs#L925-L927
https://github.com/eigen-trust/protocol/blob/master/circuit/src/gadgets/main.rs#L1023-L1025
https://github.com/eigen-trust/protocol/blob/master/circuit/src/gadgets/main.rs#L1107-L1109
[...]

PR #237, opened 2023-05-18
UnassignedValue

4rmanc:


lazovicff:
I updated the issue: https://github.com/eigen-trust/protocol/issues/207

So we want to Implement this trait for PublicKey and Signature, but not for the Circuits. Instead, the circuits will call `UnassignedPublicKey::without_witnesses` and `Signature::without_witnesses` inside its own `without_witnesses` method.

Also, it would be a good practice to do this for every test circuit

circuit/src/dynamic_sets/mod.rs
```diff
@@ -65,27 +65,23 @@ impl<const NUM_NEIGHBOURS: usize, const NUM_ITER: usize, const INITIAL_SCORE: u1
 {
 	/// Constructs a new EigenTrustSet circuit
 	pub fn new(
-		pks: Vec<PublicKey>, signatures: Vec<Signature>, op_pks: Vec<Vec<PublicKey>>,
-		ops: Vec<Vec<Scalar>>,
+		pks: Vec<UnassignedPublicKey>, signatures: Vec<UnassignedSignature>,
+		op_pks: Vec<Vec<UnassignedPublicKey>>, ops: Vec<Vec<Scalar>>,
```
lazovicff, line 69: I think it will better DX to pass normal values here, so: `Vec<PublicKey>` and `Vec<Signature>`.
Then inside this function you create the unassigned values from them.
You already created `UnassignedPublicKey::from(x)` and `UnassignedSignature::from(x)`, so this should be easy

circuit/src/eddsa/native.rs
```diff
@@ -9,13 +11,20 @@ use crate::{
...
 use rand::RngCore;
 
 type Hasher = Poseidon<Fr, 5, Params>;
 
+/// UnassignedValue Trait
+pub trait UnassignedValue {
+	/// Returns unknown value type
+	fn without_witnesses() -> Self;
+}
+
```
lazovicff, line 27: We should move this to `lib.rs` file

4rmanc:
When we want to implement Value to the test circuits, we should deal with Value type to make it work.
Do you think is [this](https://github.com/eigen-trust/protocol/blob/unassigned_value/circuit/src/ecc/mod.rs#L1057-L1067) okay?

lazovicff:
> When we want to implement `Value` to the test circuits, we should deal with `Value` type to make it work. Do you think is [this](https://github.com/eigen-trust/protocol/blob/unassigned_value/circuit/src/ecc/mod.rs#L1057-L1067) okay?

The right solution is to add `UnassignedInteger` and `UnassingedEcPoint` and implement `UnassignedValue` for them

circuit/src/ecc/native.rs
```diff
@@ -181,6 +186,46 @@ where
...
+	/// Y coordinate of the UnassignedEcPoint
+	pub y: UnassignedInteger<W, N, NUM_LIMBS, NUM_BITS, P>,
+}
+
+impl<W: FieldExt, N: FieldExt, const NUM_LIMBS: usize, const NUM_BITS: usize, P>
+	From<EcPoint<W, N, NUM_LIMBS, NUM_BITS, P>> for UnassignedEcPoint<W, N, NUM_LIMBS, NUM_BITS, P>
+where
+	P: RnsParams<W, N, NUM_LIMBS, NUM_BITS>,
+{
+	fn from(int: EcPoint<W, N, NUM_LIMBS, NUM_BITS, P>) -> Self {
```
lazovicff, line 211:
```suggestion
	fn from(ec_point: EcPoint<W, N, NUM_LIMBS, NUM_BITS, P>) -> Self {
```

circuit/src/eddsa/native.rs
```diff
@@ -91,6 +95,26 @@ impl PublicKey {
 	}
 }
 
+/// Configures a structure for the unassigned public key.
+#[derive(Clone, Copy, Default, Debug)]
+pub struct UnassignedPublicKey(pub UnassignedPoint<Fr, BabyJubJub>);
+
+impl From<PublicKey> for UnassignedPublicKey {
+	fn from(sig: PublicKey) -> Self {
```
lazovicff, line 103:
```suggestion
	fn from(pk: PublicKey) -> Self {
```

circuit/src/verifier/transcript/mod.rs
```diff
@@ -451,7 +451,7 @@ mod test {
 		type FloorPlanner = SimpleFloorPlanner;
 
 		fn without_witnesses(&self) -> Self {
-			self.clone()
+			Self { ec_point: C::default() }
```
lazovicff, line 454: We should use `UnassignedEcPoint` inside this circuit.

```rust
struct TestCommonEcPointCircuit {
  ec_point: UnassignedEcPoint,
}
```

circuit/src/verifier/transcript/mod.rs
```diff
@@ -565,7 +565,7 @@ mod test {
 		type FloorPlanner = SimpleFloorPlanner;
 
 		fn without_witnesses(&self) -> Self {
-			self.clone()
+			Self { scalar: Scalar::default() }
```
lazovicff, line 568: This should be `Value::unknown()`

circuit/src/verifier/transcript/mod.rs
```diff
@@ -647,7 +647,7 @@ mod test {
 		type FloorPlanner = SimpleFloorPlanner;
 
 		fn without_witnesses(&self) -> Self {
-			self.clone()
+			Self { reader: Vec::new() }
```
lazovicff, line 650: This should be handled the same way as:
https://github.com/eigen-trust/protocol/blob/master/circuit/src/verifier/transcript/mod.rs#L698-L714

circuit/src/verifier/transcript/mod.rs
```diff
@@ -792,7 +792,7 @@ mod test {
...
```
lazovicff, line 795: Same as:
https://github.com/eigen-trust/protocol/blob/master/circuit/src/verifier/transcript/mod.rs#L698-L714

lazovicff (approved):

PR #247, opened 2023-05-23
Update the `EigenTrustSet`(DynamicSet) to use `Attestaion`

### LINKED ISSUE #216: Adding and validating opinions inside Dynamic Sets
```rust
pub struct AttestationFr {
  pub about: Fr,
  pub key: Fr,
  pub value: Fr,
  pub message: Fr,
}

pub struct SignedAttestation {
  pub attestation: Attestation,
  pub signature: RecoverableSignature,
}

fn update_op(from: PublicKey, op: Vec<SignedAttestation>) -> Fr {
  let scores = vec![Fr::zero(); NUM_NEIGHBOURS];
  
  let mut hashes = Vec::new();
  for (i, att) in op.enumerate() {
    assert!(att.attestation.about == self.set[i]);
  
    let hash = poseidon_hash(att.attestation);
    
    if att.signature != RecoverableSignature::default() {
        let res = verify(att.attester, att.signature, hash);
        assert!(res);
    }
    scores[i] = att.attestation.value;
    
    hashes.push(hash);
  }
  
  let pos_from = self.set.iter().position(|&(x, _)| x == from);
  assert!(pos_from.is_some());
  
  self.ops.insert(from, scores);

  let op_hash = poseidon_sponge(hashes);
  return op_hash;
}
```

Since `AttestationFr` and `SignedAttestation` are going to be used for Halo2 version of Dynamic Sets, we should aim to keep them in `circuit` directory, to avoid cyclic dependencies

duguorong009:
## Description
Introduce the `Attestation` struct & update `EigenTrustSet` to use new struct
 - Closes #216 

## Changes
 - Create new file `ecdsa_native.rs` inside `circuit/dynamic-sets` dir
 - Implement the ECDSA native version of `EigenTrustSet` in new file
 - Add tests similar to EDDSA native version of `EigenTrustSet`


duguorong009:
@lazovicff 
I would like you to give more explanation on the [AttestationFr](https://github.com/eigen-trust/protocol/blob/attestation-dynamic-set/circuit/src/dynamic_sets/attest_native.rs#L42-L51) struct, especially `about` field.
I could not get the relation between `Fr` and `ECDSAPublicKey`.

lazovicff:
> ```rust
> pub fn recover_ethereum_address_from_pk(public_key: secp256k1::PublicKey) -> Address {
> 	let public_key_bytes = public_key.serialize_uncompressed();
> 	let hashed_public_key = keccak256(&public_key_bytes[1..]);
> 	let address_bytes = &hashed_public_key[hashed_public_key.len() - 20..];
> 	Address::from_slice(address_bytes)
> }
> ```

@duguorong009 UPDATE:
Instead of `Address`, we should use `Fr` as a return parameter:
```rust
pub fn recover_ethereum_address_from_pk(public_key: secp256k1::PublicKey) -> Fr {
	let public_key_bytes = public_key.serialize_uncompressed();
	let hashed_public_key = keccak256(&public_key_bytes[1..]);
	let address_bytes = &hashed_public_key[hashed_public_key.len() - 20..];
	Fr::from_bytes(address_bytes)
}
```

lazovicff:
@duguorong009 Yes, because its only 20 bytes, it will never exceed the field

circuit/src/dynamic_sets/attest_native.rs
```diff
@@ -0,0 +1,292 @@
...
+	Fr::from_bytes(&address_bytes_array).unwrap()
+}
+
+/// Dynamic set for EigenTrust
+pub struct EigenTrustAttestationSet<
+	const NUM_NEIGHBOURS: usize,
+	const NUM_ITERATIONS: usize,
+	const INITIAL_SCORE: u128,
+> {
+	set: Vec<(Option<Fr>, Fr)>,
```
lazovicff, line 39: We should not use `Option`, instead we should use Fr::zero(), for null/default values

circuit/src/dynamic_sets/attest_native.rs
```diff
@@ -0,0 +1,292 @@
...
+		let first_available = self.set.iter().position(|&(x, _)| x.is_none());
+		let index = first_available.unwrap();
+
+		// Give the initial score.
+		let initial_score = Fr::from_u128(INITIAL_SCORE);
+		self.set[index] = (Some(pk_fr), initial_score);
+	}
+
+	/// Remove the member and its opinion
+	pub fn remove_member(&mut self, pk: ECDSAPublicKey) {
```
lazovicff, line 68: `pk` should be `Fr`

circuit/src/dynamic_sets/native.rs
```diff
@@ -145,109 +156,98 @@ impl<const NUM_NEIGHBOURS: usize, const NUM_ITERATIONS: usize, const INITIAL_SCO
...
-		self.ops.insert(from, op);
-	}
+		let mut scores = vec![Fr::zero(); NUM_NEIGHBOURS];
+		let mut hashes = Vec::new();
+		for (i, att) in op.iter().enumerate() {
+			let is_default_pubkey = self.set[i].0 == Fr::zero();
+
+			if is_default_pubkey {
+				scores[i] = Fr::default();
+				hashes.push(AttestationFr::default().hash());
```
lazovicff, line 183: `AttestationFr::default().hash()` can be defined before the loop, and reused.

The rest looks good

lazovicff (approved):

PR #251, opened 2023-05-29
fix: remove `EtVerifierWrapper`

### LINKED ISSUE #233: Remove `EtVerifierWrapper`
The `EtVerifierWrapper` smart contract is outdates and has a bug in a way we process calldata.
Overall, since our strategy is focusing on off-chain, it is enough to generate a yul verifier with `snark-verifier`.

brech1:
# Related Issues

 - Resolves #233 

# Description

This PR removes the EigenTrust Verifier contract wrapper, since it will not be used for verification. Instead we will just rely on the `.yul` verifier contract and process the data off-chain.

# Changes

- Remove all `EtVerifierWrapper` related data and functions.
- Update the client configuration to manage the verifier contract instead of the wrapper.
- Update the README.
- Small changes in tests.


lazovicff:
Can you uncomment this line https://github.com/eigen-trust/protocol/blob/fix/remove-et-verifier-wrapper/data/et_verifier.yul#L1738, and then run `cargo run --release -- compile` to generate new binary

PR #252, opened 2023-05-30
feat: attestation key from domain

### LINKED ISSUE #239: Add domain ID inside client-config.json
Should be a 20 byte hex value inside the client-config.json. This value should represent the domain ID.
A prefix of 12 bytes constant value should be prepended to the ID.

Each client should only focus on one domain ID, for simplicity, so:
This prefix should serve also as a value used for indexed search of AttestationStation events.

brech1:
# Related Issues

- Resolves #239 

# Description

This PR enables the creation of the attestation 32-byte `key` value from a 12-byte pre-defined `DOMAIN_PREFIX` and a 20-byte `domain` identifier configurable by the user.

# Changes

- Adds `domain` to `client-config`.
- Updates CLI for client configuration.
- Updates attestation creation from `attest` command.
- Adds `test_attest_data_to_attestation()`.

### DIFF
--- a/client/src/attestation.rs
+++ b/client/src/attestation.rs
@@ -37,11 +37,17 @@ impl Attestation {
 		let mut about_bytes_array = [0u8; 32];
 		about_bytes_array[..about_bytes.len()].copy_from_slice(about_bytes);
 
+		let mut key = [0u8; 32];
+		self.key.to_little_endian(&mut key);
+
+		let mut message = [0u8; 32];
+		self.message.to_little_endian(&mut message);
+
 		AttestationFr {
 			about: Scalar::from_bytes(&about_bytes_array).unwrap(),
-			key: Scalar::from(self.key.0[0]),
+			key: Scalar::from_bytes(&key).unwrap(),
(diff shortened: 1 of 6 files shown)


client/src/attestation.rs
```diff
@@ -32,11 +37,17 @@ impl Attestation {
 		let mut about_bytes_array = [0u8; 32];
 		about_bytes_array[..about_bytes.len()].copy_from_slice(about_bytes);
 
+		let mut key = [0u8; 32];
+		self.key.to_little_endian(&mut key);
```
lazovicff, line 41: When constructing Fr version, we should just use the last 20 bytes.

FYI, the Scalar wouldn't be able to hold the whole 32 bytes anyway, since its limited to 254 bit field

lazovicff: So the signature will be made just on this domain ID instead of the whole key

lazovicff: We can rename `AttestationFr.key` to `AttestationFr.domain`

lazovicff (approved):

PR #253, opened 2023-06-01
generic hasher traits

4rmanc:


4rmanc:
@lazovicff Our [rescue_prime](https://github.com/eigen-trust/protocol/blob/generic_hasher/circuit/src/rescue_prime/mod.rs#L31) hasher is implemented as a `chip` and our traits are for `chipset`. Should I convert it to `chipset` with empty config or is there any better way?

lazovicff:
Use generic hasher in:
- Transcript
- Merkle Tree
- Dynamic Sets

lazovicff (approved):

PR #262, opened 2023-06-14
feat: format scores

### LINKED ISSUE #260: Writing converged scores into CSV file
peer_address - ethereum address of a peer
score_fr - final score inside the field
numerator - numerator part of integer score
denominator- denominator part of integer score
score - numerator / denominator - can be derived, but will be here for show
```csv
peer_address, score_fr, numerator, denominator, score
0x70997970C51812dc3A010C7d01b50e0d17dc79C8, 0x17942a6a1e6d480a143b3a2bc6cc20e49c4af5440c6d2da3db85036ae13cb096, 10664981617307564247061668090426822705392011863801204176349018537133273297046, 31639075521289000000000000000000000000000000000000000000000000000000000, 345
```

brech1:
# Related Issues

- Resolves #260 

# Description

This PR updates the formatting and storage of calculated scores and ensures the length of the score vectors aligns with the number of participants.

It also includes updates in the native set, to change the denominator of the rational score to `1` if it's `0`.

# Changes

- Implemented formatting of calculated scores.
- Added set and scores assertions.
- Updated the `create_csv_file` function to handle multiple lines of data.
- Included a 'scores.csv' file.
- Default the denominator to `1` if `0` in the `native_ecdsa` and `native` sets rational score calculation.


client/src/lib.rs
```diff
@@ -229,12 +229,24 @@ impl Client {
 		}
 
 		// Calculate the trust scores for each participant
-		let scores = eigen_trust_set.converge();
+		let scores = eigen_trust_set.converge_rational();
```
lazovicff, line 232: We should keep both of these.
We will have `scores_fr` and `scores_rational`

client/src/lib.rs
```diff
@@ -229,12 +229,24 @@ impl Client {
...
+
+		// Format scores to write to CSV file
+		let formatted_scores: Vec<Vec<String>> = participants
+			.iter()
+			.zip(scores.iter())
+			.map(|(participant, score)| {
+				vec![
+					format!("{:?}", participant),
+					format!("0x{:x}", score.to_integer()),
```
lazovicff, line 241: Here we should use `scores_fr`

client/src/lib.rs
```diff
@@ -229,12 +229,24 @@ impl Client {
...
+					format!("{}", score.numer()),
+					format!("{}", score.denom()),
+					format!("{}", score.numer() / score.denom()),
```
lazovicff, line 244: Here, we should use `scores_rational[i].to_integer()`

lazovicff (approved):

PR #271, opened 2023-06-19
Matrix operation

4rmanc:


circuit/src/gadgets/main.rs
```diff
@@ -685,6 +685,62 @@ impl<F: FieldExt> Chipset<F> for OrChipset<F> {
 	}
 }
 
+/// Chip for matrix operation
+pub struct MatrixChipset<F: FieldExt> {
```
lazovicff, line 689: We should rename this to `MulAddChipset`

lazovicff (approved):

PR #272, opened 2023-06-20
feat: add the constants - `POWER_OF_TEN`, `NUM_LIMBS`

### LINKED ISSUE #228: Experiment with `converge_rational` results
Currently, the `check_threshold` function accepts constants `NUM_LIMBS` and `POWER_OF_TEN`, which need to be determined by the developer beforehand. These constants depend on the `NUM_NEIGHBOURS`, `NUM_ITERATION` and the values inside the opinion of each peer.

Our goal: Given the constants `NUM_NEIGHBOURS` and `NUM_ITERATION` find the constants `NUM_LIMBS` and `POWER_OF_TEN`, such that they work for every possible value used inside the opinion matrix

duguorong009:
## Description
- Add the constants for score threshold checking - `POWER_OF_TEN` & `NUM_LIMBS`

Closes #228 

## Changes
- Add more tests in `dynamic_sets` module(native)
- Add new constants in `threshold` module(native)


circuit/src/dynamic_sets/native.rs
```diff
@@ -974,10 +968,121 @@ mod test {
...
+		let rng = &mut thread_rng();
+		let mut ops_raw = [(); NUM_NEIGHBOURS].map(|_| [(); NUM_NEIGHBOURS].map(|_| 0_u128));
+		for i in 0..NUM_NEIGHBOURS {
+			for j in 0..NUM_NEIGHBOURS {
+				ops_raw[i][j] = rng.gen::<u8>() as u128;
+			}
+			let idx = rng.gen::<usize>() % NUM_NEIGHBOURS;
+			ops_raw[i][idx] = 1;
+			ops_raw[idx][i] = 1;
+			ops_raw[i][i] = 0;
```
lazovicff, line 996: Why are we assigning 1s and 0s at these indices?

circuit/src/threshold/native.rs
```diff
@@ -27,11 +36,20 @@ impl<const NUM_LIMBS: usize, const POWER_OF_TEN: usize> Threshold<NUM_LIMBS, POW
...
-		let den_decomposed =
+		let mut den_decomposed =
 			decompose_big_decimal::<Fr, NUM_LIMBS, POWER_OF_TEN>(den.to_biguint().unwrap());
 
+		// Apply the scaling to `num` and `den`
+		let num_last_non_zero_idx = find_last_non_zero_index(&num_decomposed);
+		let den_last_non_zero_idx = find_last_non_zero_index(&den_decomposed);
+		let limb_idx = num_last_non_zero_idx
+			.unwrap_or_default()
+			.max(den_last_non_zero_idx.unwrap_or_default());
```
lazovicff, line 49: We should check if `num_last_non_zero_idx == den_last_non_zero_idx` otherwise we cant do threshold checks

lazovicff: Another constraint could be that last limb of denominator can't be zero, in any case

circuit/src/threshold/native.rs
```diff
@@ -1,10 +1,19 @@
...
 };
 use halo2::{arithmetic::Field, halo2curves::bn256::Fr};
+use num_bigint::BigUint;
 use num_rational::BigRational;
 
+/// Number of limbs for threshold check
+pub const NUM_LIMBS: usize = 55;
+/// Power of 10 for threshold check
+pub const POWER_OF_TEN: usize = 76;
+
```
lazovicff, line 16: What does this corresponds to?
4 neighbours or 128 neighbours?

lazovicff: Ah, now I realise you tried to find the constants for both of them.
You should look for different constants for each of them.

circuit/src/dynamic_sets/native.rs
duguorong009: These are for adding more randomization.
When I try the `ops_raw[i][j] = rng.gen::<u8>() as u128;` in the opinion score generation, it often outputs 2-digit numbers in most case.

circuit/src/threshold/native.rs
duguorong009: Okay, I see the point.
Mostly, the `num_last_non_zero_idx` and `den_last_non_zero_idx` are the same. 
But, there could be the case that they are different, you mean.

circuit/src/threshold/native.rs
duguorong009: Oh, so you mean that we don't want the universal constants. 
But, we need to calculate the constants for every case. (atm, 4 peers & 128 peers)
Right?

lazovicff: Yes

lazovicff (approved):

PR #275, opened 2023-06-21
Bits2Integer Chipset

4rmanc:


circuit/src/gadgets/bits2integer.rs
```diff
@@ -0,0 +1,211 @@
...
+	}
+}
+
+impl<W: FieldExt, N: FieldExt, const NUM_LIMBS: usize, const NUM_BITS: usize, P> Chipset<N>
+	for Bits2IntegerChipset<W, N, NUM_LIMBS, NUM_BITS, P>
+where
+	P: RnsParams<W, N, NUM_LIMBS, NUM_BITS>,
+{
+	type Config = Bits2IntegerChipsetConfig;
+	type Output = [Vec<AssignedCell<N, N>>; NUM_LIMBS];
```
lazovicff, line 56: We should merge all the bits here and return one big array

lazovicff: If we go with limbs, each limb should be 68 bits, and we check them individually, so it will never wrap around the field

lazovicff: > If we want to return one vector which carries the composed field value of the limbs
We are not returning composed value, instead we are returning bits that we get from Bits2Num

lazovicff (approved):

PR #276, opened 2023-06-21
feat: initial bandada integration

### LINKED ISSUE #210: Integrate with Bandada - Part 1
Add new command for adding members into semaphore groups through the Bandada API.
```
bandada --action add --id 38922764296632428858395574229367 --addr 0x70997970C51812dc3A010C7d01b50e0d17dc79C8
```

Command should do the following:
1) Load the calculated scores (from cold storage) for a particular EigenTrust set
2) Check if peer with addr is in the set
3) Check if that peer has a score more than `X` (where `X` is predefined constant) amount of reputation score inside client-config.json file.
4) Add the identity commitment to the semaphore group through the API if they 2. and 3. are true

brech1:
# Description

This PR enables adding and removing members from a semaphore group through the client CLI using the Bandada API.

## Related Issues

- Resolves #210 

## Changes

- Add `bandada` CLI command.
- Implement `bandada.rs` module to handle the api client.
- Update `.env` file management.
- Modify `read_csv_file()` and add test.
- Update readme.


lazovicff:
@brech1 Is this ready?

client/src/cli.rs
```diff
@@ -183,54 +183,30 @@ pub struct BandadaData {
...
-	type Err = &'static str;
-
-	fn from_str(s: &str) -> Result<Self, Self::Err> {
-		match s {
-			"add" => Ok(Action::Add),
-			"remove" => Ok(Action::Remove),
-			_ => Err("Invalid action."),
-		}
-	}
-}
```
lazovicff: I like this idea, why remove it?

brech1: Yes I liked it too just wasn't 100% sure about it, but if you like it then I'll add it again 😆

client/src/bandada.rs
```diff
@@ -0,0 +1,60 @@
+//! # Bandada API module.
+//!
+//! Bandada API handling module.
+
+use dotenv::{dotenv, var};
+use reqwest::header::{HeaderMap, HeaderValue, CONTENT_TYPE};
+use reqwest::{Client, Error, Response};
+
+/// Base URL for the Bandada API.
+pub const BASE_URL: &str = "http://localhost:3000"; // "https://bandada.appliedzkp.org/api"
```
lazovicff, line 10: This is a good candidate for client-config.json. wdyt?

lazovicff (approved):

PR #280, opened 2023-06-27
feat(circuit): create `AggregatorChipset`

### LINKED ISSUE #221: Turn Aggregator into Chipset
Since Aggregator will be used as a chipset in other circuits, we should implement the `Chipset` trait for it instead of keeping it as a `Circuit`.

Prerequisite for task https://github.com/eigen-trust/protocol/issues/183

duguorong009:
## Description
- Create the `AggregatorChipset` for Halo2 circuit

## Related issues
Closes #221 

## Changes
- Create `AggregatorChipset` in `circuit/verifier/aggregator/mod.rs`
- Update the unit test


circuit/src/verifier/aggregator/mod.rs
```diff
@@ -132,58 +120,14 @@ impl AggregatorConfig {
...
-
-		AggregatorConfig { common, main, poseidon_sponge, ecc_mul_scalar }
-	}
+	type Output = Vec<AssignedCell<Fr, Fr>>;
 
 	/// Synthesize the circuit.
 	fn synthesize(
-		&self, config: Self::Config, mut layouter: impl Layouter<Fr>,
-	) -> Result<(), Error> {
+		self, _common: &CommonConfig, config: &Self::Config, mut layouter: impl Layouter<Fr>,
```
lazovicff, line 129: We should use this common instead of keeping it in `AggregatorConfig`

duguorong009: Hmm, you mean that we should remove the `common` from `AggregatorConfig`, and use the above one?

lazovicff: Yes, exactly

lazovicff (approved):

PR #284, opened 2023-06-27
Batched Mul Chipset

4rmanc:
Did implementation until table select part in native implementation. 
Will find best way how to do that native part in circuit.
`p` and `scalar` types has `NUM_LIMBS` just for implementation purpose, will change it in the future commits.

### DIFF
--- a/circuit/src/ecc/same_curve/mod.rs
+++ b/circuit/src/ecc/same_curve/mod.rs
@@ -10,10 +10,11 @@ use crate::{
 		AssignedInteger, IntegerAddChip, IntegerDivChip, IntegerMulChip, IntegerReduceChip,
 		IntegerSubChip,
 	},
-	params::rns::RnsParams,
+	params::{ecc::EccParams, rns::RnsParams},
 	utils::assigned_as_bool,
 	Chip, Chipset, CommonConfig, FieldExt,
 };
+use halo2::halo2curves::ff::PrimeField;
 use halo2::{
 	circuit::{AssignedCell, Layouter},
 	halo2curves::CurveAffine,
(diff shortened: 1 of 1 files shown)


circuit/src/ecc/same_curve/mod.rs
```diff
@@ -848,6 +849,147 @@ where
...
+		let mut aux_fins: Vec<AssignedPoint<C, NUM_LIMBS, NUM_BITS, P>> = Vec::new();
+		for _ in 0..self.scalar.len() {
+			aux_fins.push(aux_fin.clone());
+			let double_chip = EccDoubleChipset::new(aux_fin);
+			aux_fin = double_chip.synthesize(
+				common,
+				&config.double,
+				layouter.namespace(|| "fin_double"),
+			)?;
+		}
```
lazovicff, line 950: Do we need to do this inside the chip. Looks like they are deterministic/same every time. Can we put them inside fixed columns?

4rmanc: You are right. I will send them from outside as we did [here](https://github.com/eigen-trust/protocol/blob/master/circuit/src/ecc/same_curve/mod.rs#L950-L1063). I will calculate these values with given `pair length` and send it to the circuit.

4rmanc:
@lazovicff I found something interesting [here](https://github.com/eigen-trust/protocol/blob/batched_mul_chipset/circuit/src/ecc/same_curve/native.rs#L515-L565) and I put it to last commit for now. I am not sure if we are aware of this but `mul_scalar` and `batched_mul_scalar` gives different results when given point is random. For example, [this](https://github.com/eigen-trust/protocol/blob/batched_mul_chipset/circuit/src/ecc/same_curve/native.rs#L542-L548) coordinates are taken from `G1Affine::random` function so they are on the curve. If we change 1 number in it `mul_scalar` and `batched_mul_scalar` gives different results. I will try to understand this more.

circuit/src/ecc/same_curve/mod.rs
4rmanc: I changed our `aux_init` and `aux_fin` [calculations](https://github.com/eigen-trust/protocol/blob/batched_mul_chipset/circuit/src/ecc/same_curve/mod.rs#L1166-L1246) in test cases, so we are calculating all of the operations outside of the chip.

circuit/src/ecc/same_curve/mod.rs
```diff
@@ -1058,7 +1217,31 @@ mod test {
...
+				aux_init = double_chip.synthesize(
+					common,
+					&config.ecc_double,
+					layouter.namespace(|| "init_double"),
+				)?;
+			}
+
+			let mut aux_fin = to_sub.clone();
+			let mut aux_fins: Vec<AssignedPoint<C, NUM_LIMBS, NUM_BITS, P>> = Vec::new();
+			for _ in 0..self.batch_length {
```
lazovicff, line 1234:
```suggestion
			let mut aux_fins: Vec<AssignedPoint<C, NUM_LIMBS, NUM_BITS, P>> = vec![aux_fin.clone()];
			for _ in 1..self.batch_length {
```

circuit/src/ecc/same_curve/mod.rs
```diff
@@ -1058,7 +1217,31 @@ mod test {
 				layouter.namespace(|| "to_sub assigner"),
 			)?;
 
-			Ok((to_add, to_sub))
+			let mut aux_init = to_add.clone();
+			let mut aux_inits: Vec<AssignedPoint<C, NUM_LIMBS, NUM_BITS, P>> = Vec::new();
+			for _ in 0..self.batch_length {
```
lazovicff, line 1222:
```suggestion
			let mut aux_inits: Vec<AssignedPoint<C, NUM_LIMBS, NUM_BITS, P>> = vec![aux_init.clone()];
			for _ in 1..self.batch_length {
```

circuit/src/ecc/same_curve/mod.rs
```diff
@@ -1013,20 +1163,29 @@ mod test {
...
 
-	struct AuxAssigner;
+	struct AuxAssigner {
+		batch_length: usize,
+		window_size: u32,
+	}
+
+	impl AuxAssigner {
+		fn new(batch_length: usize, window_size: u32) -> Self {
+			Self { batch_length, window_size }
```
lazovicff, line 1173:
```suggestion
		fn new(batch_length: usize, window_size: u32) -> Self {
		        assert!(batch_length > 0);
		        assert!(window_size > 0);
			Self { batch_length, window_size }
```

circuit/src/ecc/same_curve/mod.rs
```diff
@@ -1385,4 +1570,138 @@ mod test {
...
+			&self, config: TestConfig, mut layouter: impl Layouter<N>,
+		) -> Result<(), Error> {
+			let mut assigned_scalars_dum = [(); 10].map(|_| None);
+			layouter.assign_region(
+				|| "scalar_mul_values",
+				|region: Region<'_, N>| {
+					let mut ctx = RegionCtx::new(region, 0);
+					for i in 0..self.scalars.len() {
+						let value = ctx.assign_advice(config.common.advice[0], self.scalars[i])?;
+						assigned_scalars_dum[i] = Some(value.clone());
```
lazovicff, line 1623:
```suggestion
			let mut assigned_scalars_dum = Vec::new();
			layouter.assign_region(
				|| "scalar_mul_values",
				|region: Region<'_, N>| {
					let mut ctx = RegionCtx::new(region, 0);
					for i in 0..self.scalars.len() {
						let value = ctx.assign_advice(config.common.advice[0], self.scalars[i])?;
						assigned_scalars_dum.push(value);
```

circuit/src/ecc/same_curve/mod.rs
```diff
@@ -1058,7 +1219,31 @@ mod test {
...
+			let mut aux_fin = to_sub.clone();
+			let mut aux_fins: Vec<AssignedPoint<C, NUM_LIMBS, NUM_BITS, P>> = vec![aux_fin.clone()];
+			for _ in 1..self.batch_length {
+				aux_fins.push(aux_fin.clone());
+				let double_chip = EccDoubleChipset::new(aux_fin);
+				aux_fin = double_chip.synthesize(
+					common,
+					&config.ecc_double,
+					layouter.namespace(|| "fin_double"),
+				)?;
```
lazovicff, line 1243: Adding to array should be below the chip now.
```suggestion
				let double_chip = EccDoubleChipset::new(aux_fin);
				aux_fin = double_chip.synthesize(
					common,
					&config.ecc_double,
					layouter.namespace(|| "fin_double"),
				)?;
				aux_fins.push(aux_fin.clone());
```
Im trying to avoid calling the chip if its not needed. We will move this AuxAssigned chipset outside the tests at some point

circuit/src/ecc/same_curve/mod.rs
4rmanc: The reason I put array in here is when I use `Vec` it doubles the amount of the elements somehow. If our input length is 10 it will put 20 elements in the `Vec` and half of them will be `None`. Do you know what causes this?

lazovicff: Ah because the layouter goes through the region twice.
First time to understand the shape: https://github.com/privacy-scaling-explorations/halo2/blob/main/halo2_proofs/src/circuit/floor_planner/single_pass.rs#L93
Second time to assign: https://github.com/privacy-scaling-explorations/halo2/blob/main/halo2_proofs/src/circuit/floor_planner/single_pass.rs#L114

You should declare `assigned_scalars` inside the region itself and then return it at the end

lazovicff: We should check the rest of the crate for similar cases

lazovicff (approved):

PR #285, opened 2023-06-28
feat: storage options

### LINKED ISSUE #223: Multiple storage options
We want to make our library such that the data is passed in and passed out, and the developers using the library can chose whether. It could be:
- Database
- Memory
- Cold storage (files)

Two possible solutions:
- Simply accept and return data structures in the library itself
- Add a trait that different structs will implement depending on the storage of choice.
Example:
```rust
trait StorageHandler {
  fn read_attestations() -> Vec<Attestation>;
  fn write_attestations(atts: Vec<Attestation>);

  fn read_scores() -> Vec<PeerScore>;
  fn write_scores(atts: Vec<PeerScore>);

  fn read_threshold() -> Vec<ThresholdData>;
  fn write_threshold(atts: Vec<ThresholdData>);
}
```

Then when implementing:
```rust
pub async fn get_attestations<S: StorageHandler>(&self, st_handler: S) -> Result<_, EigenError>{
  ...
}

pub async fn calculate_scores<S: StorageHandler>(&mut self, st_handler: S) -> Result<(), EigenError>{
  ...
}
```

brech1:
# Description

This PR enables new ways of handling data storage. It adds the `Storage` trait which can be implemented on any struct that needs to add storage capabilities. It also has the `CSVFileStorage` implementation, enabling saving and loading structs on `.csv` files.

## Related Issues

- Resolves #223 

## Changes

- Implement `storage.rs` module.
- Remove `utils.rs` module.
- Adjust existing code.


client/src/storage.rs
```diff
@@ -0,0 +1,133 @@
...
+impl<T: Clone> Storage<T> for MemoryStorage<T> {
+	fn load(&self) -> Result<T, &'static str> {
+		Ok(self.data.clone())
+	}
+
+	fn save(&mut self, data: T) -> Result<(), &'static str> {
+		self.data = data;
+		Ok(())
+	}
+}
```
lazovicff, line 39: We dont need this impl

brech1:
@lazovicff Thanks for the review! I focused on the currently used storage option, being `CSVFileStorage`. Let me know what you think about this mixed approach, having a Trait implemented on different Structs for each case.

client/src/lib.rs
```diff
@@ -285,13 +277,16 @@ impl Client {
...
+
+			formatted_scores.push(record);
 		}
 
-		// Write scores to a CSV file
-		create_csv_file("scores", &formatted_scores).unwrap();
+		// Save
+		CSVFileStorage::<ScoreRecord>::new("../data/scores.csv".into())
+			.save(formatted_scores)
+			.unwrap();
```
lazovicff, line 289: The library should not hardcode specific implementation, instead we should pass a generic type parameter and call save on that function.
Also it should not know about the location of the file.

lazovicff (approved):
LGTM!

PR #289, opened 2023-06-30
feat(client): attestation records

### LINKED ISSUE #268: Save attestations to a file
In the library, we should have a separate function to return all the attestations from AS.
The binary/CLI should have a separate command to write these attestations to a file.
We also need a command for loading these attestations from a file and calculates scores.

We also have a command for doing these 2 operations together:
1) Reads the blockchain and updates the file with the latest attestations
2) Calls the calculate_scores within the same command

The format is to be decided

brech1:
# Description

This PR updates the CLI and improves the `Client` autonomy. The `calculate_scores` and `get_attestations` commands have been updated for better usability.

It also implements two new commands:

- `attestations`: Retrieves all attestations, saving them in `attestations.csv`.
- `local-scores`: Uses saved attestations to calculate scores, storing results in `scores.csv`.

And updates the `scores` command that now also retrieves and saves attestations, updating the local file.

## Related Issues

- Resolves #268 

## Changes

- `MIN_PEER_COUNT` constant introduced to the circuit.
- `AttestationRecord` created with needed conversions.
- Added `handle_attestations` and `handle_scores` functions.
- Updated `calculate_scores`. Separated `get_attestations` and `get_logs` into individual functions.
- Included an `attestations.csv` sample file.


client/src/attestation.rs
```diff
@@ -155,6 +181,93 @@ impl AttestationPayload {
...
+	value: String,
+	/// Optional field for attaching additional information to the attestation.
+	message: String,
+	/// The 'r' value of the ECDSA signature.
+	sig_r: String,
+	/// The 's' value of the ECDSA signature.
+	sig_s: String,
+	/// Recovery id of the ECDSA signature.
+	rec_id: String,
+}
```
lazovicff, line 201: Do you think storage.rs would be a better place for this, since we already keep ScoreRecord there?

brech1: I agree!

lazovicff (approved):

PR #305, opened 2023-07-06
AuxAssigner and AssignedAux

lazovicff:
Changes in this PR:
- Moved IntegerAssigner outside tests to be used by the rest of the crate
- Moved PointAssigner outside tests
- Moved AuxAssigner outside tests
- Updated Aggregator to use new AuxAssigner
- Updated Cargo.lock
- Updated clippy rules
- Updated range loops


lazovicff:
@4rmanc We should review and merge this before #304 .
We want to re-use IntegerAssigner, PointAssigner, and AuxAssigner in that PR

duguorong009 (approved):

Issue #299, opened 2023-07-03
Debug compile_contracts


lazovicff:
Investigate evm changes and solidity compiler changes

brech1:
Resolved in #307

PR #307, opened 2023-07-06
fix: `compile_sol_contracts`

brech1:
# Description

This PR replaces the `compile_sol_contracts` function with the `compile_att_station` function. The latter has a new check for compilation errors. And even though we could check for multiple contracts in that directory, `.yul` contracts would introduce errors, instead of being skipped.

Another approach to this could be to filter the `.sol` and give those paths to the compiler.

It's important to have in mind that this function will not work if the local solidity compiler version is different to those accepted by the contracts being compiled. In our case, `AttestationStation` has `pragma solidity 0.8.17;`, so the 0.8.17 version has to be the one used when running this command. This could introduce issues when dealing with multiple contracts with different versions.

## Related Issues

- #299


lazovicff (approved):
Looks good, given that we know that the only contract we will have is the AttestationStation

brech1:
> Looks good, given that we know that the only contract we will have is the AttestationStation

Awesome, then we're good to go 👍

PR #310, opened 2023-07-10
feat(circuit): implement `SelectItemChip`

### LINKED ISSUE #295: SelectItemChip
Given the index, find the item in the set (array) at this index.
Use similar technique as #293  

Example:
```rust
let item = 0
let target_item = 5;
for item in set {
  let diff = target_item - item;
  if diff == 0 {
    item = target_item;
  } else {
    item = item;
  }
}

return index;
```

duguorong009:
## Description
- Implement the halo2 chip - `SelectItemChip`, which selects the item with `index` value from a set of field values

## Related issues
Closes #295 

## Changes
- Implement the `SelectItemChip` in `circuit/gadgets/set.rs`
- Add the unit tests


circuit/src/gadgets/set.rs
```diff
@@ -279,6 +279,142 @@ impl<F: FieldExt> Chip<F> for SetPositionChip<F> {
...
+				let mut item_prev_cell = ctx.assign_from_constant(common.advice[7], F::ZERO)?;
+				ctx.next();
+
+				let mut assigned_item = ctx.assign_from_constant(common.advice[7], F::ZERO)?;
+				let mut assigned_target_idx =
+					ctx.copy_assign(common.advice[0], self.idx.clone())?;
+				for i in 0..self.items.len() {
+					ctx.enable(*selector)?;
+
+					let id_cell = ctx.assign_from_constant(common.advice[1], F::from(i as u64))?;
```
lazovicff, line 381: I think we should use fixed column for this.
Underneath, this uses both advice and a fixed column, so we can make it a little bit cheaper by just using fixed column

duguorong009: Hmm, I tried the update you recommended.
There is a small issue.
When doing the unit tests, the `production` test fails with `synthesis` error.
Normal `MockProver` test is okay.
I am digging into it.

duguorong009: I think we should use the `advice` column for `id` because it is variant which depends on `items` - circuit input.
Specifically, `id` is the column for assigning the index of `items`. 
So, I think it is variant, and `advice` column should be used.

lazovicff: It fails because a production prover goes through the circuit 2 times. First at compile time when fixed columns are assigned, then again at proving time where fixed columns are not assigned and values returned are Value::unassigned().
Since you are using this value here: https://github.com/eigen-trust/protocol/blob/select-item-chip/circuit/src/gadgets/set.rs#L381, it gives you error.

lazovicff: You can do this to solve it:
```rust
let id = F::from(i as u64);
ctx.assign_fixed(common.fixed[1], id)?;

// Calculating difference between given target idx and id of element from the set.
let diff = self.idx.value().cloned() - Value::known(id);
```

lazovicff: Or you could keep it as advice, but constrain it to be incremental (1 higher then previous row), and not use fixed column

lazovicff: Ok, but keep this in mind when we get to optimising the circuit layout at the end, in case these fixed columns stop us from having better layout

lazovicff (approved):

PR #314, opened 2023-07-12
Opinion Chipset

4rmanc:


4rmanc:
@lazovicff In the Native version at [here](https://github.com/eigen-trust/protocol/blob/master/circuit/src/opinion/native.rs#L67) we get our `msg_hash` variable from the `Poseidon Hash`. In the circuit part this is impossible to achieve because we need `Integer` type as well to create `AssignedInteger`. Is it best to change native part of the Opinion and convert attestation variables to the `Integer` type. Then get their each limb as a hash input and at the end of limb hashing 4 times we can hash that 4 output again with Poseidon to get single output. Do you agree with this method or will it be very costly?

lazovicff:
> @lazovicff In the Native version at [here](https://github.com/eigen-trust/protocol/blob/master/circuit/src/opinion/native.rs#L67) we get our `msg_hash` variable from the `Poseidon Hash`. In the circuit part this is impossible to achieve because we need `Integer` type as well to create `AssignedInteger`. Is it best to change native part of the Opinion and convert attestation variables to the `Integer` type. Then get their each limb as a hash input and at the end of limb hashing 4 times we can hash that 4 output again with Poseidon to get single output. Do you agree with this method or will it be very costly?

The other solution could be to pass 2 versions of msg_hash. First one to be a field element `F` and the other `Integer`, then check: compose(msg_hash_int) == msg_hash_f, then take the integer and work with it?

4rmanc:
Where exactly you mean? We cannot pass `msg_hash` because we are getting it from the hash result.

lazovicff:
Ok so `mgs_hash_f` will be calculated inside, but `msh_hash_int` will be passed through the constructor, or more precisely a vector of them.
Turning attestation values to integers and hashing them would add more overhead since we would need to use sponges

circuit/src/opinion/mod.rs
```diff
@@ -279,11 +288,11 @@ where
 					|| "constraint equality",
 					|region: Region<'_, N>| {
 						let mut ctx = RegionCtx::new(region, 0);
-						ctx.assign_advice(
+						let msg_hash = ctx.assign_advice(
 							common.advice[0],
 							Value::known(P::compose(self.msg_hash[i].integer.limbs)),
```
lazovicff, line 293: This is not enough, we have to compose it using MainGate.

4rmanc: You mean we need a new chipset which calculates composed value of the given `AssignedCell` limbs?
We need composed value of the `AssignedPublicKey` as well. In that case I won't change the structure of it for now.

lazovicff: We can use `MulAddChipset` to replicate this logic:
```rust
let left_shifters = Self::left_shifters();
let mut sum = N::ZERO;
for i in 0..NUM_LIMBS {
	sum = MulAddChipset::new(input[i], left_shifters[i], sum).synthesize(...);
}
sum
```
Where `left_shifters` are assigned to fixed columns.

For public keys, lets use the limbs for hashing. I think the collision attacks are possible if we compose them

4rmanc: Okay, I got it and I added TODO for it and I will do it after the meeting.

4rmanc:
@lazovicff FYI, especially in the test part I did some additions manually to test if native and circuit returns the same result. For example, I added [public_key](https://github.com/eigen-trust/protocol/blob/opinion_chipset/circuit/src/opinion/mod.rs#L840) manually like this. Assertions and the other parts seems like working fine.

lazovicff (approved):

PR #316, opened 2023-07-13
feat(circuit): implement `ThresholdCircuit`

### LINKED ISSUE #234: Threshold Circuit
Make a separate `Circuit` implementation for `Threshold`. 

It should be the halo2 version of the native algorithm for threshold checks.

The steps are:
- Verify DynamicSets proof (Using aggregator) SKIP for now.
- Select a specific Address from a set using SetPositionChip (pass address as PI)
- Take the score associated with this address (Using SelectItem chip)
- Do threshold check on this score

Anonymous threshold checks (NOTE: Skip for now):
- Verify DynamicSets proof (Using aggregator)
- Using the secret key construct the public key, and then hash the key to construct the address (using keccak)
- Constrain that address is in the set
- Find the score at the same index as the address
- Do threshold check on this score

duguorong009:
## Description
- Implement the `ThresholdCircuit` - halo2 version of threshold checks algorithm

## Related issues
Closes #234 

## Changes
- Implement the `ThresholdCircuit` in `circuit/threshold/mod.rs`
- Add the unit tests


circuit/src/threshold/mod.rs
```diff
@@ -1,2 +1,649 @@
...
+
+		// max_score = NUM_NEIGHBOURS * INITIAL_SCORE
+		let max_score = {
+			let mul_chipset = MulChipset::new(num_neighbor, init_score);
+			mul_chipset.synthesize(
+				&config.common,
+				&config.main,
+				layouter.namespace(|| "NUM_NEIGHBOURS * INITIAL_SCORE"),
+			)?
+		};
```
lazovicff, line 253: No need for new scope because `synthesize` will drop `mul_chipset`

circuit/src/threshold/mod.rs
```diff
@@ -1,2 +1,649 @@
...
+
+		let composed_den_inv = {
+			let inv_chipset = InverseChipset::new(composed_den);
+			inv_chipset.synthesize(
+				&config.common,
+				&config.main,
+				layouter.namespace(|| "composed_den ^ -1"),
+			)?
+		};
```
lazovicff, line 403: Also here the new scope is not needed

circuit/src/threshold/mod.rs
duguorong009: Done https://github.com/eigen-trust/protocol/pull/316/commits/6fb493afaea8d9b5e40a7b86460dbb329bc149d0

circuit/src/threshold/mod.rs
duguorong009: Done https://github.com/eigen-trust/protocol/pull/316/commits/6fb493afaea8d9b5e40a7b86460dbb329bc149d0

circuit/src/threshold/mod.rs
```diff
@@ -1,2 +1,645 @@
...
+						config.common.instance,
+						inst_col_offset + i,
+					)?;
+					sets_pk_y.push(member);
+
+					if i % ADVICE == ADVICE - 1 {
+						ctx.next();
+					}
+				}
+				inst_col_offset += NUM_NEIGHBOURS;
```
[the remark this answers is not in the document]
duguorong009: I followed the pattern(+ example) of `eddsa` version of `EigenTrustSet` circuit, not the `ecdsa` version.
That's why I used the `public keys`, not `addresses`.

So, do you think we should use the `address` here, assuming that `ecdsa` version? @lazovicff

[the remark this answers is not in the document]
duguorong009: Done https://github.com/eigen-trust/protocol/pull/316/commits/b7ef601b2abcb952baa28bef16365f8b41c66f00

duguorong009:
> We should rethink the private/public input structure:
> 
> * set - private because we will pass it to aggregator
> * scores - private because we will pass it to aggregator
> * target_address - public
> * threshold - public
> 
> Few things to note:
> 
> * `score` should not be passed at all since we will get it from `scores` at some index
> * It is possible that prover can pass invalid `set` and `scores`, but we will solve it with attestation validity checks, by hashing all the attestations at a specific order inside the circuit and check the hash against some public input
> * Private inputs should be passed through the constructor
> * Public inputs should be passed to the verifier

okay, thanks for the detail!
Will refactor the circuit as recommended!

lazovicff (approved):

PR #319, opened 2023-07-17
feat(client): standard errors

### LINKED ISSUE #297: Standardise Error for the Client
Use EigenError across the crate

brech1:
# Description

This PR aims to implement the usage of the `EigenError` variants across the whole client.

## Related Issues

- Resolves #297 

# Changes

- Add `thiserror` dependency.
- Update error types.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -986,6 +986,7 @@ dependencies = [
  "secp256k1 0.27.0",
  "serde",
  "serde_json",
+ "thiserror",
  "tokio",
 ]
 
--- a/client/Cargo.toml
+++ b/client/Cargo.toml
@@ -19,6 +19,7 @@ reqwest = "0.11.18"
 secp256k1 = { version = "0.27.0", features = ["global-context", "recovery"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
+thiserror = "1.0.43"
 tokio = { version = "1.18", features = ["time", "macros", "rt-multi-thread", "net"] }
 
 eigen-trust-circuit = { path = "../circuit" }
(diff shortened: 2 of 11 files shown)


lazovicff:
There are cases where we can propagate EigenError, instead of unwrapping.
I will list the ones I found:
https://github.com/eigen-trust/protocol/blob/feat/use-eigenerror/client/src/cli.rs#L285
https://github.com/eigen-trust/protocol/blob/feat/use-eigenerror/client/src/cli.rs#L294

https://github.com/eigen-trust/protocol/blob/feat/use-eigenerror/client/src/lib.rs#L227
https://github.com/eigen-trust/protocol/blob/feat/use-eigenerror/client/src/lib.rs#L231
https://github.com/eigen-trust/protocol/blob/feat/use-eigenerror/client/src/lib.rs#L256
https://github.com/eigen-trust/protocol/blob/feat/use-eigenerror/client/src/lib.rs#L274

There are other unwraps that can also be turned into EigenError, wdyt?

brech1:
> There are cases where we can propagate EigenError, instead of unwrapping. I will list the ones I found: https://github.com/eigen-trust/protocol/blob/feat/use-eigenerror/client/src/cli.rs#L285 https://github.com/eigen-trust/protocol/blob/feat/use-eigenerror/client/src/cli.rs#L294
> 
> https://github.com/eigen-trust/protocol/blob/feat/use-eigenerror/client/src/lib.rs#L227 https://github.com/eigen-trust/protocol/blob/feat/use-eigenerror/client/src/lib.rs#L231 https://github.com/eigen-trust/protocol/blob/feat/use-eigenerror/client/src/lib.rs#L256 https://github.com/eigen-trust/protocol/blob/feat/use-eigenerror/client/src/lib.rs#L274
> 
> There are other unwraps that can also be turned into EigenError, wdyt?

Thanks for the review! I certainly left some unwraps that should have been removed. 

Totally, we could remove every unwrap from the client and map the error to an EigenError, but perhaps is better to implement it into a different PR since it will incur in having multiple functions/methods now returning `Result<_,EigenError>`. But let me know your thoughts.

lazovicff (approved):

PR #323, opened 2023-07-21
chore(zk): remove the codes of generating yul verifiers

duguorong009:
## Description
- Remove the code of generating yul verifiers
- Remove the unit tests which are using the above verifiers

## Related issues
Closes #300 

## Changes
- Remove the `gen_evm_verifier`, `gen_evm_verifier_code` and `evm_verify` from `zk/src/verifier` module
- Remove the unit tests which are using `evm_verify` util function

### DIFF
--- a/zk/src/verifier/mod.rs
+++ b/zk/src/verifier/mod.rs
@@ -1,12 +1,12 @@
 use halo2::{
 	dev::MockProver,
 	halo2curves::{
-		bn256::{Bn256, Fq, Fr, G1Affine},
+		bn256::{Bn256, Fr, G1Affine},
 		group::ff::PrimeField,
 	},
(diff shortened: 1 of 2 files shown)


zk/src/verifier/mod.rs
```diff
@@ -82,65 +74,12 @@ pub fn gen_proof<C: Circuit<Fr>>(
...
-	let result = evm.call_raw(caller, verifier_address, calldata.into(), 0.into());
-
-	dbg!(result.gas_used);
-	dbg!(result.reverted);
-	dbg!(result.exit_reason);
-
-	let success = !result.reverted;
-	assert!(success);
-}
-
```
lazovicff, line 135: Lets keep this, maybe we will need it at some point

lazovicff (approved):

PR #325, opened 2023-07-25
DynamicSets Update

4rmanc:


eigentrust-zk/src/circuits/dynamic_sets/mod.rs
```diff
@@ -514,80 +514,52 @@ impl<const NUM_NEIGHBOURS: usize, const NUM_ITER: usize, const INITIAL_SCORE: u1
...
-					let op_pk_j_y = op_pk_y_i[j].clone();
 
-					// Condition 1. op_pk_j != pk_i
-					let equal_chip = IsEqualChipset::new(op_pk_j_x.clone(), pk_i_x.clone());
-					let is_pk_i_x = equal_chip.synthesize(
+				for j in 0..NUM_NEIGHBOURS {
+					let addr_j = set[j].clone();
+					// Condition 1. addr_j != addr_i
+					let equal_chip = IsEqualChipset::new(addr_j.clone(), addr_i.clone());
+					let is_add_i = equal_chip.synthesize(
```
duguorong009, line 522: I think the name should be `is_addr_i`, not the `is_add_i`.
right? @4rmanc @lazovicff

eigentrust-zk/src/circuits/dynamic_sets/mod.rs
```diff
@@ -514,80 +514,52 @@ impl<const NUM_NEIGHBOURS: usize, const NUM_ITER: usize, const INITIAL_SCORE: u1
...
-					let and_chip = AndChipset::new(is_pk_i_x, is_pk_i_y);
-					let is_pk_i = and_chip.synthesize(
-						&config.common,
-						&config.main,
-						layouter.namespace(|| "op_pk_j == pk_i"),
-					)?;
-					let sub_chip = SubChipset::new(one.clone(), is_pk_i);
-					let is_diff_pk = sub_chip.synthesize(
+					let sub = SubChipset::new(one.clone(), is_add_i);
+					let is_not_add_i = sub.synthesize(
```
duguorong009, line 528: This one also should be `is_not_addr_i`, I think.

eigentrust-zk/src/verifier/aggregator/mod.rs
```diff
@@ -194,6 +197,7 @@ impl Chipset<Fr> for AggregatorChipset {
 					G1Affine,
 					_,
 					Bn256_4_68,
+					5,
```
duguorong009, line 200: Same as above.(Please use `WIDTH`, not `5`)

eigentrust-zk/src/verifier/aggregator/mod.rs
```diff
@@ -213,6 +217,7 @@ impl Chipset<Fr> for AggregatorChipset {
 				G1Affine,
 				_,
 				Bn256_4_68,
+				5,
```
duguorong009, line 220: Same as above. (Please use `WIDTH`, not `5`)

eigentrust/src/attestation.rs
```diff
@@ -33,6 +37,11 @@ pub type ECDSAPublicKey = secp256k1::PublicKey;
 pub type ECDSASignature = ecdsa::RecoverableSignature;
 /// Signature represented with field elements
 pub type SignatureFr = Signature<Secp256k1Affine, Scalar, NUM_LIMBS, NUM_BITS, Secp256k1_4_68>;
+/// Attestation represented with field
+pub type AttestationFr = Attestation<Scalar>;
```
duguorong009, line 41: I think we should either rename the `AttestationFr` as `AttestationScalar`, or rename the `Attestation<Scalar>` as `Attestation<Fr>`.
Personally, I would go for the `AttestationScalar`.
I believe it is better for future reading.

duguorong009 (commented):
In the codebase, we are switching to the use of `Scalar` or generic name `N` for the places of `halo2_curves::bn256::Fr`.
Eventually, we should remove all of `Fr`, and use the `Scalar` or `N` for better reading.
Maybe, this could be done in separate PR, in future.

lazovicff:
Thanks for the review @duguorong009 . Can you check again if I got all of them?

duguorong009:
> Thanks for the review @duguorong009 . Can you check again if I got all of them?

Yes, I will re-check the PR. 
Thanks for the quick fix!

duguorong009 (approved):

Issue #224, opened 2023-05-17
Split library and CLI


lazovicff:
Split lib.rs and its modules into a different crate.
Explore different naming conventions for both the cli and the library

brech1:
Solved in #326

PR #326, opened 2023-07-25
restructure: split client and cli

brech1:
# Description

This PR splits the eigentrust library from the CLI application.

# Related Issues

- #224 

# Changes

- Add `cli` crate.
- Remove environment file usage from the library.
- Update client constructor.


Cargo.toml
```diff
@@ -1,6 +1,7 @@
 [workspace]
 members = [
-	"zk",
-	"client",
+	"eigentrust",
+	"eigentrust-cli",
+	"eigentrust-zk",
 ]
-default-members = ["client"]
+default-members = ["eigentrust-cli"]
```
lazovicff, line 7: How do we choose a default?

lazovicff:
`.gitignore` should be updated to ignore params

brech1:
> `.gitignore` should be updated to ignore params

Thanks, I forgot to update the path. Params are ignored with the `/eigentrust/assets/params-*.bin` rule, but I didn't delete them, should I remove them all params?

lazovicff:
> > `.gitignore` should be updated to ignore params
> 
> Thanks, I forgot to update the path. Params are ignored with the `/eigentrust/assets/params-*.bin` rule, but I didn't delete them, should I remove all params?

Yes, we can remove them

lazovicff (approved):

PR #337, opened 2023-07-29
feat(lib): standard client interface

### LINKED ISSUE #308: Standardize Client Outputs
Values passed in and returned from Client methods should be in standardised form, namely in raw bytes.

Few more points:
- `AttestationCreatedFilter` is not a good choice for attestations because it requires post-processing, and contains ethereum related types, instead `get_attestations` return type should contain both Attestation and AttestationPayload values
- `calculate_scores` should accept this new type that contains Attestation and AttestationPayload values
- `calculate_scores` should return a type that contains (`peer_address`, `score_fr`, `score_ratio`) as raw bytes

brech1:
# Description

This PR standardizes the inputs and outputs of the main eigentrust library methods to have a more raw interface in order to avoid unnecessary pre/post processing for the data.

## Related Issues

- Resolves #308 

## Changes

- Remove unused rand crate from the main lib.
- Implement conversions from the contract logs to raw data structs.
- Update storage and cli to handle new data types.
- Fix tests.

### DIFF
--- a/eigentrust/src/lib.rs
+++ b/eigentrust/src/lib.rs
@@ -58,14 +58,13 @@ use crate::attestation::{SignatureEth, SignatureRaw, SignedAttestationEth};
 use att_station::{
 	AttestationCreatedFilter, AttestationData as ContractAttestationData, AttestationStation,
 };
-use attestation::AttestationEth;
+use attestation::{AttestationEth, AttestationRaw, SignedAttestationRaw};
 use eigentrust_zk::{
 	circuits::dynamic_sets::ecdsa_native::{
-		EigenTrustSet, RationalScore, SignedAttestation as SignedAttestationFr, MIN_PEER_COUNT,
-		NUM_BITS, NUM_LIMBS,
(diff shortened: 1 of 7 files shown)


eigentrust/src/lib.rs
```diff
@@ -120,10 +121,17 @@ pub struct ClientConfig {
...
+/// Score struct.
+pub struct Score {
+	/// Participant address.
+	pub address: [u8; 20],
+	/// Scalar score.
+	pub score_fr: [u8; 32],
+	/// Rational score (numerator, denominator).
+	pub score_rat: ([u8; 32], [u8; 32]),
+	/// Hexadecimal score.
+	pub score_hex: [u8; 32],
```
lazovicff, line 133: How are `score_fr` and `score_hex` different?

brech1: - `score_hex` is the truncated integer representation of the `BigRational` (return of `converge_rational()`) expressed as a hex string. It could also be a String.
- `score_fr` is the scalar score, return value of `converge()`.

But it might be confusing, should I change the names of them? Or add some extra clarification maybe

lazovicff (approved):

PR #338, opened 2023-07-29
Ecdsa assigner

4rmanc:


lazovicff:
We should pass both `s` and `s_inv` and check if `s * s_inv == 1` inside either Ecdsa or EcdsaAssigner.
I think doing it inside the Ecdsa chipset is more correct

eigentrust-zk/src/ecdsa/mod.rs
```diff
@@ -431,17 +432,191 @@ where
...
+				let assign_res = ctx.assign_advice(
+					common.advice[0],
+					Value::known(
+						<P as RnsParams<C::Scalar, N, NUM_LIMBS, NUM_BITS>>::compose(
+							is_one.result.limbs,
+						),
+					),
+				)?;
+				let one = ctx.assign_fixed(common.fixed[0], N::ONE)?;
+				ctx.constrain_equal(assign_res, one)?;
```
lazovicff, line 594: Here we are just assigning a number to `advice[0]` and a number to `fixed[0]` and checking if they are the same, but we are not using these numbers anywhere.

The correct solution is to pass `s: UnassignedInteger` and `s_inv: UnassignedInteger` and then:
1) Make AssignedInteger's from them
2) Multiply them using `IntegerMulChip`
3) Assign AssignedInteger that holds value of 1
3) Constrain the AssignedInteger from step 2. to be equal to AssignedInteger from step 3.

4rmanc: Okay, done. Somehow I was thinking `mul()` from native is the circuit multiplication. Thank you.

PR #339, opened 2023-07-31
fix(zk): use `Aggregator` in `Threshold` circuit

### LINKED ISSUE #318: Use aggregator inside Threshold Circuit
1. Aggregate 1 DynamicSets proof inside the Threshold circuit, by passing `set` and `scores` as instances
2. Find the index of target address inside the `set`
3. Get the score from the found index
4. Check if the score is above threshold

duguorong009:
## Description
- Introduce the `AggregatorChipset` inside the `ThresholdCircuit`

## Related issues
Closes #318 

## Changes
- Use the `Aggregator` in `ThresholdCircuit`, in order to validate the `sets` & `scores`
- Update the unit test

### DIFF
--- a/eigentrust-zk/src/params/hasher/poseidon_bn254_5x5.rs
+++ b/eigentrust-zk/src/params/hasher/poseidon_bn254_5x5.rs
@@ -1,4 +1,5 @@
 use super::*;
+use halo2::halo2curves::bn256::Fr;
 
 #[derive(Clone)]
 /// Configures a structure.
@@ -25,7 +26,7 @@ impl Sbox for Params {
 	}
 }
 
-impl<F: FieldExt> RoundParams<F, 5> for Params {
+impl RoundParams<Fr, 5> for Params {
 	fn partial_rounds() -> usize {
 		60
 	}
--- a/eigentrust-zk/src/params/hasher/poseidon_bn254_5x5.rs
+++ b/eigentrust-zk/src/params/hasher/poseidon_bn254_5x5.rs
@@ -1,5 +1,4 @@
 use super::*;
-use halo2::halo2curves::bn256::Fr;
 
 #[derive(Clone)]
 /// Configures a structure.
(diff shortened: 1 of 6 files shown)


eigentrust-zk/src/params/hasher/poseidon_bn254_5x5.rs
```diff
@@ -26,7 +25,7 @@ impl Sbox for Params {
 	}
 }
 
-impl RoundParams<Fr, 5> for Params {
+impl<F: FieldExt> RoundParams<F, 5> for Params {
```
lazovicff, line 28: Why is this necessary?

duguorong009: In order to use the `Aggregator` inside `Threshold` circuit, we need the `AggregatorConfig`.
This config needs the `FullRoundChip selector` and `PartialRoundChip Selector`, from `Poseidon` hasher.(https://github.com/eigen-trust/protocol/blob/374d7fd4dcf82863f5ee9318414dddb342887a85/eigentrust-zk/src/circuits/threshold/mod.rs#L172-L173)
Here, we need the `RoundParams` generic parameter.
Originally, I was going to use the `Params(poseidon_bn254_5*5.rs)` directly, not using the generic parameter.
That's why I updated the definition, so that it can use generic paramter, not specific `Bn256::Fr` field.

Now, I updated the `Threshold` circuit in order to introduce more generic parameters.(https://github.com/eigen-trust/protocol/blob/374d7fd4dcf82863f5ee9318414dddb342887a85/eigentrust-zk/src/circuits/threshold/mod.rs#L42-L66)
Hence, we don't need above change.
I will roll back the changes in `hasher::params` dir.

duguorong009:
I think I over-engineered this PR. 
I will create new PR for same purpose.

lazovicff:
@duguorong009 I was reviewing this and I think you were on the right path.
Which part did you over-engineer?

duguorong009:
> @duguorong009 I was reviewing this and I think you were on the right path. Which part did you over-engineer?

Oh, sorry for inconvenience.
I found that I should update all other `verifier`-related code, including `gen_pk`, etc.
And it needs much refactoring(generic params ...).

I thought I can just use the `bn256::Fq`, `bn256::Fr` directly. 
So, we can wrap up the whole flow at the moment.(Atm, @4rmanc  fixed the Aggregator issue)

Issue #345, opened 2023-08-21
Use the native Threshold circuit for Bandada checks


lazovicff:
On this line:
https://github.com/eigen-trust/protocol/blob/master/eigentrust-cli/src/cli.rs#L234

We are just taking the score and doing normal comparison check.
We want to use the Threshold native circuit: https://github.com/eigen-trust/protocol/blob/master/eigentrust-zk/src/circuits/threshold/native.rs

`score: F, ratio: BigRational, threshold: F` are the values loaded from the file

The properties of the struct are going to be useful for the circuit public and private inputs: https://github.com/eigen-trust/protocol/blob/master/eigentrust-zk/src/circuits/threshold/native.rs#L18-L21

This function: https://github.com/eigen-trust/protocol/blob/master/eigentrust-zk/src/circuits/threshold/native.rs#L60-L96 should replace the comparison check

brech1:
Resolved in #347

PR #347, opened 2023-08-30
feat: native threshold circuit

brech1:
# Description

This PR aims removes the comparison check for the bandada group score threshold from the `cli` crate and implements the `Threshold` circuit to verify the score value.

## Related Issues

- #345 

## Changes

- Add `num_rational` dependency to the `eigentrust` lib to create a `BigRational`.
- Parse the score values loaded from the file and implement the native `Threshold` circuit to validate the score.

### DIFF
--- a/eigentrust/src/lib.rs
+++ b/eigentrust/src/lib.rs
@@ -62,8 +62,8 @@ use att_station::{
 use attestation::{AttestationEth, AttestationRaw, SignedAttestationRaw};
 use eigentrust_zk::{
 	circuits::{
-		dynamic_sets::native::EigenTrustSet, PoseidonNativeHasher, PoseidonNativeSponge,
-		HASHER_WIDTH, MIN_PEER_COUNT, NUM_BITS, NUM_LIMBS,
+		dynamic_sets::native::EigenTrustSet, threshold::native::Threshold, PoseidonNativeHasher,
+		PoseidonNativeSponge, HASHER_WIDTH, MIN_PEER_COUNT, NUM_BITS, NUM_LIMBS,
 	},
 	ecdsa::native::{EcdsaKeypair, PublicKey, Signature},
 	halo2::halo2curves::{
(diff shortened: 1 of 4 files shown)


eigentrust/src/lib.rs
```diff
@@ -410,6 +412,27 @@ impl Client {
...
+
+	/// Verifies if a participant's score surpasses the bandada score threshold.
+	pub fn verify_threshold(
+		score: u64, score_num: u64, score_den: u64, threshold: u64,
+	) -> Result<(), EigenError> {
+		let score_fr = Scalar::from(score);
+		let threshold_fr = Scalar::from(threshold);
+		let score_ratio = BigRational::new(score_num.into(), score_den.into());
+
+		let th_circuit: Threshold<Scalar, NUM_LIMBS, POWER_OF_TEN, MAX_NEIGHBOURS, INITIAL_SCORE> =
```
lazovicff, line 424: Instead of `NUM_LIMBS`, we should use `NUM_DECIMAL_LIMBS` constant

eigentrust/src/lib.rs
```diff
@@ -410,6 +412,27 @@ impl Client {
...
+
+		if th_circuit.check_threshold() {
+			info!("Participant score passes the bandada score threshold.");
+			Ok(())
+		} else {
+			Err(EigenError::ValidationError(
+				"Failed to check participant score against bandada score threshold.".to_string(),
+			))
+		}
+	}
```
lazovicff, line 435: Lets exclude the wording related to bandada, it should be general-purpose threshold check.

Also, when the `check_threshold` returns `false`, I dont see it as error. It simply didn't pass the check, so we should just return the bool.

In the bandada command, however, returning errors makes more since in case of failure

brech1:
Thanks for the review, I thought the threshold only applied to bandada groups, but after this review and today's meeting I understand it's part of the post processing.

lazovicff (approved):

PR #348, opened 2023-09-03
Aggregator Fix

4rmanc:
Fixed native aggregator with infinity checks and added infinity check circuit to the circuit part


eigentrust-zk/src/ecc/generic/mod.rs
```diff
@@ -129,6 +129,66 @@ where
 	) -> AssignedEcPoint<C, N, NUM_LIMBS, NUM_BITS, P> {
 		Self { x, y }
 	}
+
+	/// Checks if given point is at the infinity or not
+	pub fn is_infinity(&self) -> bool {
+		self.x.integer == Integer::zero() && self.y.integer == Integer::zero()
```
duguorong009, line 135: I think the projective coordinates of point at infinity is `(0, 1, 0)`. (https://zcash.github.io/halo2/background/curves.html#projective-coordinates)
But, here, it seems checking `(0, 0)`. 
Why?

duguorong009 (approved):

lazovicff (approved):

PR #349, opened 2023-09-04
fix(zk): use `Aggregator`  in `Threshold` circuit

### LINKED ISSUE #318: Use aggregator inside Threshold Circuit

duguorong009:
## Description
- Introduce the `AggregatorChipset` inside the `ThresholdCircuit`

## Related issues
- Close #318 

## Changes
- Use the `Aggregator` in `ThresholdCircuit`, in order to validate the `sets` & `scores`
- Update the unit test


duguorong009:
@lazovicff 
Here, I am creating a PR with simple solution.
I think this simple solution is better fit for the original goal of the issue.

As we discussed in the meeting, I make this PR as `draft` one.
So, you can review the PR, and give the feedback.
Once we complete the `Debug Aggregator` issue, I will come back, merge the updates, and make the PR `ready for review`.

lazovicff (approved):

PR #352, opened 2023-09-07
fix(lib): scores calculation

### LINKED ISSUE #350: Fix scores calculation
The scores calculation is currently failing.

```
thread 'main' panicked at 'assertion failed: pos_from.is_some()', /.../eigentrust/eigentrust-zk/src/circuits/opinion/native.rs:67:9
```

This needs to be fixed before running integration tests.

brech1:
# Description

This PR fixes the scores calculation method from the main eigentrust lib.

I also skipped the assertion for domain and about fields for empty attestations in the native opinion validation.

## Related Issues

 - Resolves #350 

## Changes

- Fix address generation and recovery endianness.
- Fix scores parsing.
- Fix opinion validation.


brech1:
@lazovicff Please notice that the scores file is generated with the score values as hex, due to the new standard interface from the client, should these values keep being stored as decimals?

eigentrust-zk/src/circuits/opinion/native.rs
```diff
@@ -72,8 +72,11 @@ where
 		let mut hashes = Vec::new();
 		for i in 0..NUM_NEIGHBOURS {
 			let att = self.attestations[i].clone();
-			assert!(att.attestation.about == set[i]);
-			assert!(att.attestation.domain == self.domain);
+
+			if att != SignedAttestation::empty(self.domain) {
+				assert!(att.attestation.about == set[i]);
+				assert!(att.attestation.domain == self.domain);
+			}
```
lazovicff, line 79: We can't have this since we want the logic to match the circuit, and the logic should be that all attestations should pass this check:
```rust
assert!(att.attestation.about == set[i]);
assert!(att.attestation.domain == self.domain);
```

The scores are then nullified in case the signature verification fails.

So, we should do the same here. When we are constructing the `Attestation::empty`, the about address should be passed as well (should be taken from the set). This logic should be done here:
https://github.com/eigen-trust/protocol/blob/fix/scores-calc/eigentrust-zk/src/circuits/dynamic_sets/native.rs#L198-L199

lazovicff:
> @lazovicff Please notice that the scores file is generated with the score values as hex, due to the new standard interface from the client, should these values keep being stored as decimals?

I was hoping they could be decimal for human readability

brech1:
I fixed the scores formatting and the `update_op` command. It now keeps track of the peer being attested in order to create the empty attestation with the correct about.

lazovicff (approved):

PR #358, opened 2023-09-21
feat(lib): attestation filtering

### LINKED ISSUE #353: Filter the attestation logs by PREFIX + config.domain

brech1:
# Description

This PR introduces changes to filter the attestations retrieved from the contract, in order to only process the ones belonging to the an EigenTrust related project and also to the configured domain.

## Related Issues

- Resolves #353 

## Changes

- Update get_logs library method.
- Add test for get_logs.

### DIFF
--- a/eigentrust/src/lib.rs
+++ b/eigentrust/src/lib.rs
@@ -55,16 +55,13 @@ pub mod eth;
 pub mod storage;
 
 use crate::{
-	attestation::{
-		SignatureEth, SignatureRaw, SignedAttestationEth, SignedAttestationScalar, DOMAIN_PREFIX,
-		DOMAIN_PREFIX_LEN,
-	},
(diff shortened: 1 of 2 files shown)


eigentrust/src/lib.rs
```diff
@@ -456,16 +462,29 @@ impl Client {
...
+		let logs = self
+			.signer
+			.get_logs(&filter)
+			.await
+			.map_err(|e| EigenError::ParsingError(e.to_string()))?;
+
+		// Filter logs based on the domain prefix.
+		let filtered_logs: Vec<Log> = logs
+			.iter()
+			.filter(|log| &log.topics[3].as_fixed_bytes()[..DOMAIN_PREFIX_LEN] == DOMAIN_PREFIX)
```
lazovicff, line 483: What about the domain id? We should also filter based on that

brech1: Yes! I'm adding that as well

lazovicff (approved):

PR #361, opened 2023-09-26
feat(lib): threshold circuit integration

### LINKED ISSUE #331: Generate Threshold proofs
- Generate params and save them to assets folder of the library
- Load the params, load the EigenTrust convergence proof and calculate Threshold proof for a given domain
- Save the proof to the json file named after the domain id and the address of the peer for which the proof is calculated for

brech1:
# Description

This PR integrates the threshold circuit in the library and enable threshold proofs generation through the cli.

## Related Issues

- Resolves #331 

## Changes 

- Create new commands to handle threshold proofs.
- Fetch attestations only if local file isn't found.
- Update file generation.
- Move `big_to_fe_rat` from test mod to actual util.
- Refactor main library to have specific functions to build the circuit structures that can be reused in multiple methods, `th_circuit_setup` and `et_circuit_setup`.

### DIFF
--- a/eigentrust-cli/src/cli.rs
+++ b/eigentrust-cli/src/cli.rs
@@ -547,10 +547,17 @@ pub async fn load_or_fetch_attestations(
 ) -> Result<Vec<SignedAttestationRaw>, EigenError> {
 	let att_file_path = get_file_path("attestations", FileType::Csv)?;
 	let att_storage = CSVFileStorage::<AttestationRecord>::new(att_file_path.clone());
-	let local_records = att_storage.load()?;
 
-	if !local_records.is_empty() {
-		return local_records.into_iter().map(|record| record.try_into()).collect();
+	match att_storage.load() {
+		Ok(local_records) => {
+			if !local_records.is_empty() {
+				return local_records.into_iter().map(|record| record.try_into()).collect();
+			}
+			debug!("No local attestations found. Fetching from AS contract.");
+		},
+		Err(_) => {
+			debug!("Error loading local attestations. Fetching from AS contract.");
+		},
 	}
 
 	let client = Client::new(config, load_mnemonic());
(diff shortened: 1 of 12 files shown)


eigentrust-cli/src/cli.rs
```diff
@@ -429,6 +432,87 @@ pub async fn handle_scores(
...
+	let mnemonic = load_mnemonic();
+	let client = Client::new(config, mnemonic);
+
+	// Load KZG params
+	let et_kzg_params = EigenFile::KzgParams(ET_PARAMS_K).load()?;
+	let th_kzg_params = EigenFile::KzgParams(TH_PARAMS_K).load()?;
+
+	// Get attestations
+	let att_fp = get_file_path("attestations", FileType::Csv)?;
+	handle_attestations(client.get_config().clone()).await?;
```
lazovicff, line 446: I think we should default to using local attestations instead of calling the chain every time.

eigentrust-cli/src/cli.rs
```diff
@@ -429,6 +432,87 @@ pub async fn handle_scores(
...
+	let client = Client::new(config.clone(), mnemonic);
+
+	// Load KZG params and proving key
+	let et_kzg_params = EigenFile::KzgParams(ET_PARAMS_K).load()?;
+	let th_kzg_params = EigenFile::KzgParams(TH_PARAMS_K).load()?;
+	let proving_key = EigenFile::ProvingKey(Circuit::Threshold).load()?;
+
+	// Get attestations
+	let att_fp = get_file_path("attestations", FileType::Csv)?;
+	handle_attestations(client.get_config().clone()).await?;
```
lazovicff, line 468: Same here

eigentrust-cli/src/cli.rs
lazovicff: Same case for these functions:
- handle_et_proof
- handle_th_proof
- handle_th_pk

lazovicff: That is also an option. It is easier to use since we wont need anvil running all the time, and we can use the same data to experiment with it

lazovicff (approved):

Issue #362, opened 2023-09-26
Extract native Integer operations from IntegerChipset


lazovicff:
During the witness generating time, the Integers inside IntegerChipsets (https://github.com/eigen-trust/protocol/blob/master/eigentrust-zk/src/integer/mod.rs#L32-L39) are initialised with 0, which means when the prover goes through our code, there are bugs associated with things like calculating `.inverse()` of a `0` or doing arithmetic with points of infinity.

To solve this - the integers should be separated from the chipsets and we should find better ways to compute `ReductionWitness`

Issue #327, opened 2023-07-27
Debug Aggregator


lazovicff:
After previously debugging, we understood that MSMs bases have points of infinity:
```
Msm { constant: None, scalars: [0x0000000000000000000000000000000000000000000000000000000000000001], bases: [Infinity] },
Msm { constant: None, scalars: [0x0000000000000000000000000000000000000000000000000000000000000001], bases: [Infinity] },
Msm { constant: None, scalars: [0x0000000000000000000000000000000000000000000000000000000000000001], bases: [Infinity] },
Msm { constant: None, scalars: [0x0000000000000000000000000000000000000000000000000000000000000001], bases: [Infinity] },
Msm { constant: None, scalars: [0x0000000000000000000000000000000000000000000000000000000000000001], bases: [Infinity] },
Msm { constant: None, scalars: [0x0000000000000000000000000000000000000000000000000000000000000001], bases: [Infinity] },
```

Our `lhs` values from the Accumulator have inconsistent values between native and halo2 version, as a consequence of that

PR #374, opened 2023-10-13
Debug aggregator(update integer chipsets)

duguorong009:
Related issues: #362 #327

### DIFF
--- a/eigentrust-zk/src/verifier/transcript/mod.rs
+++ b/eigentrust-zk/src/verifier/transcript/mod.rs
@@ -296,9 +296,9 @@ where
 		};
 
 		let assigned_integer_x =
-			AssignedInteger::<_, _, NUM_LIMBS, NUM_BITS, P>::new(x, assigned_x.try_into().unwrap());
+			AssignedInteger::<_, _, NUM_LIMBS, NUM_BITS, P>::new(assigned_x.try_into().unwrap());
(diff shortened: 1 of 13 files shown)


eigentrust-zk/src/verifier/transcript/mod.rs
```diff
@@ -132,14 +143,19 @@ where
 		self.state.update(&ec_point.inner.x.limbs);
 		self.state.update(&ec_point.inner.y.limbs);
 
+		self.native_state.update(&ec_point.inner.x.integer.limbs);
+		self.native_state.update(&ec_point.inner.y.integer.limbs);
```
lazovicff, line 147: Why do we need native_state?

duguorong009: The reason is that we need native scalar value for this op - `TranscriptReadChipset::squeeze_challenge`.
https://github.com/eigen-trust/protocol/pull/374/files#diff-11b415885c31eaa45e328fccdb132c3f01ff7015130d93ca21e7c7b929e03315R120-R136

I added the native scalar value as one field into the `Halo2LScalar` struct.
So, the above `squeeze_challenge` should have a way to get the native scalar value for `Halo2LScalar`.

lazovicff (approved):

4rmanc (approved):

